PR #37, opened 2025-08-20
Replace usage of corev1.ResourceList with map[corev1.ResourceName]int64

thiyyakat:

Replaces use of `corev1.ResourceList` with `map[corev1.ResourceName]int64` in `PodInfo`, `NodeInfo`, `PodResourceInfo` and `NodeResourceInfo`.

Fixes #

**Release note**:

```other developer

```


elankath (approved):
thanks

PR #39, opened 2025-08-22
Change types to introduce data required for NodeScore and NodeScoreSelector

thiyyakat:

The PR introduces the following:
1. a field `ScaledNodeResource` of type `NodeResourceInfo` to `NodeScore`. This allows the `LeastCost` strategy to access the `Allocatable` of the scaled node.
2. a field `Region` of type `string` to `NodePlacementInfo`. This is required to get the price of an instance based on region in the scoring logic.

Fixes #

**Release note**:

```other developer

```


elankath (approved):
/lgtm

PR #40, opened 2025-08-26
Implement scoring module

thiyyakat:

The PR implements the `least-cost` and `least-waste` strategies for node scoring. It also implements two node score selection strategies- `max-allocatable` and `min-price`, to pick between identical node score values. The PR contains corresponding tests for the good-path.

Fixes #

**Release note**:

```feature developer

```


elankath:
thanks. will merge this with maddy on fri. might need some changes.

PR #66, opened 2025-09-21
[TEST] Refactor minkapi server tests

takoverflow:
Refactor minkapi `server` tests into a uniform, readable format

Fixes #

**Release note**:

```feature user

```

### DIFF
--- a/minkapi/server/store/store_test.go
+++ b/minkapi/server/store/store_test.go
@@ -7,6 +7,7 @@ package store
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"reflect"
 	"strconv"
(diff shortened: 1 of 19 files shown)


minkapi/server/store/store_test.go
```diff
@@ -7,6 +7,7 @@ package store
 import (
```
elankath: There is already a `server_test.go` in `main`. This PR does not appear merged with `main` ?

takoverflow: This is rebased on top of `main`, it's not creating a new `server_test`, just adding the refactored minkapi server tests to the existing file.

PR #68, opened 2025-09-30
Add pull-request-target types labeled

unmarshall:
To enable GHA to checkout a PR branch it is not sufficient to only have
```yaml
on:
  pull_request_target:
```
As per https://gardener.github.io/cc-utils/github_actions.html#example-configuration-for-label-based-trust one needs to add `labeled` so that it can react on `reviewed/ok-to-test` label on a PR.

Fixes #

**Release note**:

```feature user
None
```

### DIFF
--- a/.github/workflows/non-release.yaml
+++ b/.github/workflows/non-release.yaml
@@ -6,6 +6,8 @@ on:
       - 'hotfix-**'
       - 'release-**'
   pull_request_target:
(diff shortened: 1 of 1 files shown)


elankath (approved):
/lgtm - ignoring docu lint failure for this.

takoverflow (approved):

PR #69, opened 2025-09-30
ci: add pr write permission to all jobs running trusted-checkout

takoverflow:
As per the docs, PR write permission is required for all workflows that call `trusted-checkout`
to allow for removing the `trusted-label` (note the **caveat** below)
```
         pull-requests: write # needed so trusted-checkout can remove trusted-label
                              # caveat: also needs to be set for all called workflows
                              # that use trusted-checkout (action)
```

Fixes #

**Release note**:

```feature user

```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -24,6 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
+      pull-requests: write
     steps:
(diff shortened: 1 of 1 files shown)


takoverflow:
In-lieu of https://sap-ti.slack.com/archives/C0170EXLULE/p1759244070767319?thread_ts=1759222003.667639&cid=C0170EXLULE

unmarshall (approved):

PR #78, opened 2025-11-07
fix(genscenario): Sanitize and cleanup cluster sensitive information

takoverflow:
This PR sanitizes the cluster snapshot objects to get rid of cluster specific information.
It also adds `omitempty` tags to API objects to skip `null` fields.

Fixes #

There are still references to pod names in labels, ownerRefs etc that needs to be tackled before this PR can be reviewed/merged.

**Release note**:

```feature user

```

### DIFF
--- a/api/core/v1alpha1/scalingconstraints.go
+++ b/api/core/v1alpha1/scalingconstraints.go
@@ -78,24 +78,24 @@ type NodePool struct {
 	// Region is the name of the region.
 	Region string `json:"region"`
 	// Priority is the priority of the node pool.
-	Priority int32 `json:"priority"`
+	Priority int32 `json:"priority,omitempty"`
 	// Labels is a map of key/value pairs for labels applied to all the nodes in this node pool.
-	Labels map[string]string `json:"labels"`
+	Labels map[string]string `json:"labels,omitempty"`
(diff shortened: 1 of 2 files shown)


api/service/types.go
```diff
@@ -318,16 +318,16 @@ func (n *NodeInfo) GetResourceInfo() NodeResourceInfo {
 // ResourceMeta contains common metadata fields for Kubernetes resources.
 type ResourceMeta struct {
 	// UID is the unique identifier for the resource.
-	UID types.UID
+	UID types.UID `json:"uid,omitempty"`
```
unmarshall: Should not be `omitempty`

api/core/v1alpha1/scalingconstraints.go
```diff
@@ -108,7 +108,7 @@ type NodeTemplate struct {
 	// InstanceType is the instance type of the node template.
 	InstanceType string `json:"instanceType"`
 	// Priority is the priority of the node template. The lower the number, the higher the priority.
-	Priority int32 `json:"priority"`
+	Priority int32 `json:"priority,omitempty"`
```
unmarshall: Should not be omitempty

PR #146, opened 2026-02-18
ScaleOut Plans for Pods with bound, unbound PersistentVolumeClaims and all VolumeBindingModes

### LINKED ISSUE #90: Revise ClusterSnapshot and associated API for Volumes support

Currently, the scaling advisor scale out planner does not support volume bindings by the scheduler. This needs the API for Cluster Snapshot and some other associated types to be revised.

**Why is this needed**:

Scheduler skips volume bindings while making scheduling decisions. This needs to be corrected. The basic scenarios are also being adjusted to reflect cases with volumes.

elankath:

Fixes #90

**Release note**:

```feature user
ScaleOut Plans for Pods with bound and unbound PersistentVolumeClaims
```


api/core/v1alpha1/scalingconstraints.go
```diff
@@ -84,6 +107,23 @@ type NodePool struct {
 	Priority int32 `json:"priority"`
 }
 
+// GetNodePlacements computes and returns all the possible NodePlacement's for this NodePool.
```
takoverflow:
```suggestion
// GetNodePlacements computes and returns all the possible NodePlacements for this NodePool.
```

api/planner/scaleout_types.go
```diff
@@ -0,0 +1,193 @@
...
+	Error error `json:"error,omitempty"`
+	// Labels is the associated metadata.
+	Labels map[string]string `json:"labels,omitempty"`
+	// ScaleOutPlan is the generated scale-out plan.
+	ScaleOutPlan *sacorev1alpha1.ScaleOutPlan `json:"scaleOutPlan,omitempty"`
+}
+
+// ScaleOutSimulation represents a simulation that scales virtual node(s) and performs valid unscheduled pod to ready node
+// assignments against a minkapi View.
+// A ScaleOutSimulation implementation may use a k8s scheduler - either embedded or external to do this, or it may form a SAT/MIP model
```
takoverflow: > or it may form a SAT/MIP model from the pod/node data and run a tool that solves the model

Is this something in plan for future, post MVP? What would be the scenarios where this might be needed?

takoverflow (changes requested):
Thanks for the changes!
I've not gone through the entire PR, only minor things that stood out at a first glance. PTAL

planner/testutil/testutil.go
```diff
@@ -0,0 +1,333 @@
...
+	"time"
+
+	"github.com/gardener/scaling-advisor/planner/scheduler"
+
+	commontypes "github.com/gardener/scaling-advisor/api/common/types"
+	sacorev1alpha1 "github.com/gardener/scaling-advisor/api/core/v1alpha1"
+	"github.com/gardener/scaling-advisor/api/minkapi"
+	plannerapi "github.com/gardener/scaling-advisor/api/planner"
+	"github.com/gardener/scaling-advisor/common/podutil"
+	commontestutil "github.com/gardener/scaling-advisor/common/testutil"
```
takoverflow: Since there are now multiple `testutil` packages in different modules

Can the other imports of `common/testutil` be also changed to use `commontestutil`? In `minkapi` module and `scheduler_test` of the `planner`. For uniformity and better clarity.

planner/testutil/testutil.go
```diff
@@ -0,0 +1,333 @@
...
+		return false
+	}
+	d.Request.CreationTime = time.Now()
+	d.Request.DiagnosticVerbosity = DefaultPlannerTestVerbosity
+	d.Request.ID = t.Name()
+	if args.NodeScoringStrategy != "" {
+		d.Request.ScoringStrategy = args.NodeScoringStrategy
+	} else {
+		d.Request.ScoringStrategy = commontypes.NodeScoringStrategyLeastCost
+	}
```
takoverflow:
```suggestion
	d.Request.ScoringStrategy = cmp.Or(args.NodeScoringStrategy, commontypes.NodeScoringStrategyLeastCost)
```

Similarly this can be done for other fields where we take the argument if non-nil or else fallback to the default. To decrease verbosity and have less conditional clauses, WDYT?

planner/testutil/testutil.go
```diff
@@ -0,0 +1,333 @@
...
+		return
+	}
+	constraintGenInput := samples.ConstraintGenInput{GenDir: testData.GenDir, PoolPreset: args.PoolPreset, PoolZones: args.PoolZones}
+	constraintGenOutput, err := samples.GenScalingConstraints(constraintGenInput)
+	if err != nil {
+		t.Fatalf("failed to generate constraints: %v", err)
+		return
+	}
+	testData.Request.Constraint = &constraintGenOutput.Constraint
+	for c, n := range args.NumUnscheduledPerResourcePreset {
```
takoverflow: Consider renaming `c` to `category`, I had to look at the definition to recall what `c` would be referring to.

samples/samples.go
```diff
@@ -97,26 +113,238 @@ func GenerateSimplePodsWithTemplateData(num int, podTmplData SimplePodTemplateDa
...
 }
 
-// GenerateSimplePodsForResourceCategory generates simple pods with a container specifying requests for the given resourceCategory and using the given metadata.
+// GenerateSimplePodsForResourcePreset generates simple pods with a container specifying requests for the given resourceCategory and using the given metadata.
 // Also generates the pod YAML's for these pods within the temp directory.
-func GenerateSimplePodsForResourceCategory(category ResourceCategory, num int, metadata SimplePodMetadata) (pods []corev1.Pod, podYAMLPaths []string, err error) {
-	podTmplData := SimplePodTemplateData{
-		SimplePodMetadata: metadata,
-		Resources:         category.AsResourceList(),
+func GenerateSimplePodsForResourcePreset(category ResourcePreset, num int, metadata PodGenInput) (pods []corev1.Pod, podYAMLPaths []string, err error) {
```
takoverflow: Is there a need for having this function and `GenerateSimplePodsWithTemplateData` as two separate functions? Can the latter not be subsumed into this function itself, its not being called anywhere else.

samples/types.go
```diff
@@ -65,44 +65,107 @@ type AppLabels struct {
...
+	// StorageClassName is used to set the generated PVC.spec.storageClassName and PV.spec.storageClassName
+	StorageClassName string
+	// VolumeBindingMode is used to determine whether the PVC is bound to the PV.
+	//
+	//  If VolumeBindingMode is set to "Immediate" (default), then the PV.spec.claimRef is set to refer to its associated
+	// PVC, the "pv.kubernetes.io/provisioned-by" is set on the PV, the PV.status.phase is set to "Bound", the
+	// "pv.kubernetes.io/bind-completed" annotation is set to "yes"` on PVC, the PVC.spec.volumeName refers to the PV.Spec.CSI.VolumeHandle
+	// and the PVC.status.phase is set to "Bound".
+	//
+	//  If VolumeBindingMode is set to "WaitForFirstConsumer, the above claimRef and annotations are not set, the PV.status.phase is set to "Available"
```
takoverflow:
```suggestion
	// If VolumeBindingMode is set to "Immediate" (default): 
	//   1. PV.spec.claimRef is set to refer to its associated PVC
	//   2. PV.status.phase is set to "Bound"
	//   3. "pv.kubernetes.io/provisioned-by" annotation is set on the PV
	//   4. PVC.spec.volumeName refers to the PV.Spec.CSI.VolumeHandle
	//   5. PVC.status.phase is set to "Bound".
	//   6. "pv.kubernetes.io/bind-completed" annotation is set to "yes" on PVC
	//
	// If VolumeBindingMode is set to "WaitForFirstConsumer", the above claimRef and annotations
	// are not set and the PV.status.phase is set to "Available"
```

takoverflow (changes requested):
A couple more minor comments, will need one more round to finish the review.

planner/planner.go
```diff
@@ -131,3 +127,28 @@ func wrapPlanContext(ctx context.Context, traceLogsDir string, req *plannerapi.R
...
+	if args.SchedulerLauncher == nil {
+		return fmt.Errorf("%w: schedulerLauncher must be set", plannerapi.ErrCreatePlanner)
+	}
+	if args.StorageMetaAccess == nil {
+		return fmt.Errorf("%w: storageMetaAccess must be set", plannerapi.ErrCreatePlanner)
+	}
+	if args.SimulatorFactory == nil {
+		return fmt.Errorf("%w: simulatorFactory must be set", plannerapi.ErrCreatePlanner)
+	}
+	if args.SimulationFactory == nil {
```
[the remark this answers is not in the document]
aaronfern: They're different 🤔 
SimulatorFactory vs SimulationFactory

planner/util/util.go
```diff
@@ -105,31 +166,140 @@ func groupNodeScoresByNodePlacement(nodeScores []plannerapi.NodeScore) map[sacor
...
+			}
+			sc = scObj.(*storagev1.StorageClass)
+		}
+
+		var selectedPV corev1.PersistentVolume
+		for _, pv := range pvs {
+			if _, alreadyBound := boundPVs[pv.Name]; alreadyBound {
+				continue
+			}
+			if !IsPVBindCandidate(ctx, &pvc, &pv, sc.Name) {
```
takoverflow: Is this how the actual PV -> PVC binding is done, pick the first matching candidate that satisfies the requested capacity?

If not can it cause discrepancies with regards to what the simulation might be binding from what the actual binding ends up being.

That is, the results from embedded kube-scheduler might differ from what the real scheduler would do if the volume bindings differ? Or is that not a realistic concern.

planner/util/util.go
```diff
@@ -105,31 +166,140 @@ func groupNodeScoresByNodePlacement(nodeScores []plannerapi.NodeScore) map[sacor
...
+				continue
+			}
+			selectedPV = pv
+			break
```
takoverflow: This can be simplified:
```suggestion
			if IsPVBindCandidate(ctx, &pvc, &pv, sc.Name) {
				selectedPV = pv
			    break
			}
			
```

api/core/v1alpha1/scalingadvice.go
```diff
@@ -44,7 +44,7 @@ type ScalingAdviceSpec struct {
 	// ScaleInPlan is the plan for scaling in across node pools.
 	ScaleInPlan *ScaleInPlan `json:"scaleInPlan"`
 	// ConstraintRef is a reference to the ScalingConstraint that this advice is based on.
-	ConstraintRef apicommon.ConstraintReference `json:"constraintRef"`
+	ConstraintRef apicommon.NamespacedName `json:"constraintRef"`
```
r4mek: `json:"namespacedName"`?

planner/planner_test.go
```diff
@@ -5,356 +5,253 @@
...
 	}
-	gotPlan := obtainScaleOutPlan(t, planner, &testData)
-	assertExactScaleOutPlan(t, wantPlan, gotPlan)
+	testutil.ObtainAndAssertScaleOutPlan(t, planner, &testData, wantPlan)
 }
 
-func assertExactScaleOutPlan(t *testing.T, want, got *sacorev1alpha1.ScaleOutPlan) {
-	if got == nil {
-		t.Fatalf("got nil ScaleOutPlan, want not nil ScaleOutPlan")
+// TestOnePoolHalfFitPodScaleout tests scale out of one pool using HalfBerry pods that half-fit into pool P's NodeTemplate.
```
takoverflow:
```suggestion
// TestOnePoolHalfFitPodScaleout tests scale out of one pool using HalfBerry pods that half-fit into pool A's NodeTemplate.
```

planner/planner_test.go
```diff
@@ -5,356 +5,253 @@
...
-// createPlannerAndTest creates the test ScalingPlanner and TestData for the given TestArgs.
-func createTestPlannerAndTestData(t *testing.T, args TestArgs) (planner plannerapi.ScalingPlanner, testData TestData, ok bool) {
-	if len(args.NumUnscheduledPerResourceCategory) == 0 {
-		t.Fatal("args.NumUnscheduledPerResourceCategory mandatory")
-		return
-	}
-	var err error
-	testData.RunContext, planner, ok = createTestScalingPlanner(t, args.Timeout)
+// TestOnePoolHalfAndFullFitPodScaleout tests scale out of one pool using both HalfBerry and Berry pods that half-fit
+// and full-fit into pool P's NodeTemplate.
```
takoverflow:
```suggestion
// and full-fit into pool A's NodeTemplate.
```

takoverflow (changes requested):
There were a bunch of queries where I might not be correct in my assumptions so I've requested further clarification.

Most of the comments are with regards to the discrepancy that would be there when performing `PVC<->PV` binding in the tests versus what would be done for the actual scale out simulations where there wouldn't be dummy volume related objects created and no provisioner/pv controller running.

Apologies for the slow review.

samples/samples.go
elankath: The idea is that one is a convenience function which applies a known resource preset and the other can be used if you don't want to use the known resource preset.  Technically, will have some unit-tests in the future that don't use the preset - negative tests which will directly use `GenerateSimplePodsWithTemplateData`

api/common/types/types.go
```diff
@@ -54,12 +56,31 @@ type QPSBurst struct {
...
+type NamespacedName struct {
 	// Name is the name of the ClusterScalingConstraint.
 	Name string `json:"name"`
 	// Namespace is the namespace of the ClusterScalingConstraint.
-	Namespace string `json:"namespace"`
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// AsObjectName converts this namespaced name to a client-go cache.ObjectName
+func (nn *NamespacedName) AsObjectName() cache.ObjectName {
```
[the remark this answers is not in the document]
elankath: Oversight. Changing to value receivers.

api/common/types/types.go
```diff
@@ -54,12 +56,31 @@ type QPSBurst struct {
...
+	if nn.Namespace == "" && nn.Name == "" {
+		return nil
+	} else {
+		return &corev1.ObjectReference{Namespace: nn.Namespace, Name: nn.Name}
+	}
+}
+
+// String returns the general purpose string representation
+func (nn NamespacedName) String() string {
+	return nn.Namespace + "/" + nn.Name
```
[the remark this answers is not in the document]
elankath: No, this follows the behaviour of `types.NamespacedName` - which we unfortunately want to use but cannot use, but the great k8s overlord maintainers decided not to give this type json tags for some strange reason. And they gave the crazy recommendation that every k8s project should have its own `NamespacedName`. So there are a billion `NamespacedName` across k8s now.


```go

type NamespacedName struct {
	Namespace string
	Name      string
}

const (
	Separator = '/'
)

// String returns the general purpose string representation
func (n NamespacedName) String() string {
	return n.Namespace + string(Separator) + n.Name
}

```
I am changing the Godoc and make use of `Separator` to clarify the intent.

```go
// NamespacedName is a fully qualified object name.
// NOTE: This is only needed since k8s APIMachinery types.NamespacedName does not have JSON tags and k8s maintainers
// recommended that every project should use their own copy of NamespacedName.
type NamespacedName struct {
	// Name is the name of the object.
	Name string `json:"name"`
	// Namespace is the namespace of the object.
	Namespace string `json:"namespace,omitempty"`
}

// AsObjectName converts this namespaced name to a client-go cache.ObjectName
func (nn NamespacedName) AsObjectName() cache.ObjectName {
	return cache.ObjectName{Name: nn.Name, Namespace: nn.Namespace}
}

// AsObjectReference constructs an ObjectReference referring this name or nil if name is empty.
func (nn NamespacedName) AsObjectReference() *corev1.ObjectReference {
	if nn.Name == "" {
		return nil
	} else {
		return &corev1.ObjectReference{Namespace: nn.Namespace, Name: nn.Name}
	}
}

// String returns the general purpose string representation
// Matches implementation in APIMachinery types.NamespacedName
func (nn NamespacedName) String() string {
	return nn.Namespace + string(types.Separator) + nn.Name
}
```

planner/util/util.go
```diff
@@ -105,31 +166,140 @@ func groupNodeScoresByNodePlacement(nodeScores []plannerapi.NodeScore) map[sacor
...
+		log.Error(err, "failed to bind pvc->pv", "pvc", pvc, "pv", pv)
+		return fmt.Errorf("%w: failed to bind pvc %q ->pv %q: %w", plannerapi.ErrBindClaimVolume, pvc.Name, pv.Name, err)
+	}
+
+	log.V(3).Info("bound pvc<->pv", "pvcName", pvc.Name, "pvName", pv.Name)
 	return nil
 }
+
+// IsPVBindCandidate checks whether the given PV can be selected as a bindable candidate for the given PVC.
+func IsPVBindCandidate(ctx context.Context, pvc *corev1.PersistentVolumeClaim, pv *corev1.PersistentVolume, storageClassName string) bool {
```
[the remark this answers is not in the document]
elankath: Yes, I assumed that, but you are right that assumption maybe broken in the future. Best to not assume this. `VolumeMode` is irrelevant to the actual scheduling though - it is only relevant for binding.

Will correct this and I think we also need a few unit test cases for `BindClaimsAndVolumes`. Will introduce them.

planner/util/util.go
elankath: Hmm...thinking...frankly which specific PV belonging to a zone gets bound doesn't really matter since we are not giving PV's in our ScaleOut plan. As long as a matching PV zonal unit is consumed. But I think it might make sense to avoid confusion in diagnostics and debugging to do the exact thing in pv_helpers. I will adopt the same approach to reduce confusion.

samples/samples.go
```diff
@@ -97,26 +113,238 @@ func GenerateSimplePodsWithTemplateData(num int, podTmplData SimplePodTemplateDa
...
+	}
+	for _, pvcName := range vi.PVCNames {
+		var (
+			pv          corev1.PersistentVolume
+			volumePhase = corev1.VolumeBound
+			pvName      string
+			yamlPath    string
+		)
+		if vi.VolumeBindingMode == storagev1.VolumeBindingWaitForFirstConsumer {
+			volumePhase = corev1.VolumeAvailable
```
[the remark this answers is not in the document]
elankath: Yes, adding this to godoc for `GeneratePersistentVolumes` to clarify intent.

```
// GeneratePersistentVolumes generates a slice of PersistentVolume objects bound to the given pvcNames suitable for the given provider in the given
// namespace for the given storage and access mode and returns the PV objects and their generated YAML paths.
// If StorageVolGenInput.VolumeBindingMode is `storagev1.VolumeBindingWaitForFirstConsumer`, then the generated PVs are unbound with no `claimRef` set
// and have their phase set to `corev1.VolumeAvailable`.
// Otherwise, their phase is set to `corev1.VolumeBound` with their `claimRef` referring to the PVC.

```

samples/samples.go
```diff
@@ -97,26 +113,238 @@ func GenerateSimplePodsWithTemplateData(num int, podTmplData SimplePodTemplateDa
...
+	}
+	root, err := os.OpenRoot("/")
+	if err != nil {
+		return fmt.Errorf("%w: failed to open root FS: %w", commonerrors.ErrExecuteTemplate, err)
+	}
+	return objutil.LoadIntoRuntimeObj(root.FS(), strings.TrimPrefix(outPath, "/"), obj)
+}
+
+// GetMaxAllocatableVolumes returns a reasonable default value for the max number of allocatable volumes for a node of the given instanceType.
+func GetMaxAllocatableVolumes(provider commontypes.CloudProvider, instanceType string) int32 {
```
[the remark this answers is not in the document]
elankath: This will be given when we have a PR for the service - which will be actor started outside the unit-tests and used productively and can be launched via CLI and consumed by the operator.

Node specific volume limits are already possibly via CSINode objects.

api/core/v1alpha1/scalingconstraints.go
```diff
@@ -51,6 +54,26 @@ type ScalingConstraintSpec struct {
 	NodePools []NodePool `json:"nodePools,omitempty"`
 }
 
+// GetAllNodePlacements computes all the possible NodePlacements for the ScalingConstraintSpec.
+func (c *ScalingConstraintSpec) GetAllNodePlacements() []NodePlacement {
+	placements := make([]NodePlacement, 0, len(c.NodePools))
```
unmarshall: [Minor]: Since `p.GetNodePlacements` can return a much larger number as compared to `len(c.NodePools)` there is no benefit here for pre-allocation.

This function is only used in test. Is there a need to make it part of the API? When we do use it in main code then we can have it, right?

api/core/v1alpha1/scalingconstraints.go
```diff
@@ -51,6 +54,26 @@ type ScalingConstraintSpec struct {
...
+	for _, p := range c.NodePools {
+		placements = append(placements, p.GetNodePlacements()...)
+	}
+	return placements
+}
+
+// GetAllAvailabilityZones gets all the availability zones across all node pools as a sorted slice.
+func (c *ScalingConstraintSpec) GetAllAvailabilityZones() []string {
```
unmarshall: This function is only used in test. Is there a need to make it part of the API? When we do use it in main code then we can have it, right?

api/common/types/types.go
```diff
@@ -54,12 +57,34 @@ type QPSBurst struct {
...
-	Namespace string `json:"namespace"`
+}
+
+// AsObjectName converts this namespaced name to a client-go cache.ObjectName
+func (nn NamespacedName) AsObjectName() cache.ObjectName {
+	return cache.ObjectName{Name: nn.Name, Namespace: nn.Namespace}
+}
+
+// AsObjectReference constructs an ObjectReference referring this name or nil if name is empty.
+func (nn NamespacedName) AsObjectReference() *corev1.ObjectReference {
```
unmarshall: You can simplify this to:
```go
func (nn NamespacedName) AsObjectReference() *corev1.ObjectReference {
	if nn.Name == "" {
		return nil
	}
	return &corev1.ObjectReference{Namespace: nn.Namespace, Name: nn.Name}
}
```

api/planner/errors.go
```diff
@@ -14,42 +14,49 @@ import (
...
+	ErrCreateSimulator = errors.New("cannot create simulator")
+	// ErrCreateSimulation is a sentinel error indicating that the planner cannot create a scaling simulation
+	ErrCreateSimulation = errors.New("cannot create simulation")
+	// ErrRunSimulation is a sentinel error indicating that planner could not successfully run a specific scaling simulation
+	ErrRunSimulation = errors.New("cannot run simulation")
+
+	// ErrRunSimulationGroup is a sentinel error indicating that the planner could not run a scaling simulation group.
+	ErrRunSimulationGroup = errors.New("cannot run simulation group")
+
+	// ErrBindClaimVolume is a sentinel erorr indicating that a scaling simulation cannot bind PVC<->PV
```
unmarshall:
```suggestion
	// ErrBindClaimVolume is a sentinel error indicating that a scaling simulation cannot bind PVC<->PV
```

api/planner/scaleout_types.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// SimulatorFactory is a factory facade for constructing various kinds of simulators.
+type SimulatorFactory interface {
+	GetScaleOutSimulator(args SimulatorArgs) (ScaleOutSimulator, error)
+	// TODO: Add GetScaleInSimulator here.
+}
+
+// SimulationFactory is a factory facade for creating Simulation objects
+type SimulationFactory interface {
```
unmarshall: Is this interface really required?

unmarshall (commented):
1/n reviews.

minkapi/viewutil/viewutil.go
```diff
@@ -41,27 +82,78 @@ func LogNodeAndPodNames(ctx context.Context, prefix string, view minkapi.View) e
...
-			log.V(4).Info(prefix+"|node in view",
-				"viewName", view.GetName(),
-				"nodeName", node.Name,
-				"nodePool", node.Labels[commonconstants.LabelNodePoolName],
-				"instanceType", node.Labels[corev1.LabelInstanceTypeStable],
-				"region", node.Labels[corev1.LabelTopologyRegion],
-				"zone", node.Labels[corev1.LabelTopologyZone],
-				"allocatable", node.Status.Allocatable)
-		}
+	for idx, pod := range allPods {
```
unmarshall: This is inefficient. If the log verbosity level is set to lower than 4 it still unnecessarily loops over all pods, all nodes. Instead you should explicitly check:

```go
if log.V(4).Enabled() {
   // iterate over pods, nodes, pvcs, pvs
}
```

unmarshall (commented):
2/n reviews

common/volutil/volutil.go
```diff
@@ -0,0 +1,442 @@
...
+//   - annotations["pv.kubernetes.io/bind-completed"] = "yes"
+//
+// PV must have:
+//   - spec.claimRef populated
+//   - status.phase = Bound
+func BindClaimsForImmediateMode(ctx context.Context, view minkapi.View) ([]plannerapi.VolumeClaimAssignment, error) {
+	log := logr.FromContextOrDiscard(ctx)
+	scs, pvcs, pvs, err := viewutil.ListStorageClassesClaimsAndVolumes(ctx, view)
+	if err != nil {
+		return nil, err
```
unmarshall: does it make sense to wrap the error here to provide more context that the list failed and that caused the binding of Claims for the immediate mode to fail.

common/volutil/volutil.go
```diff
@@ -0,0 +1,442 @@
...
+			sc = FindStorageClassWithName(*pvc.Spec.StorageClassName, scs)
+			if sc == nil {
+				log.V(2).Info("cannot find PVC storage class", "pvcName", pvc.Name, "pvcNamespace", pvc.Namespace, "storageClassName", *pvc.Spec.StorageClassName)
+				continue
+			}
+		}
+		if sc.VolumeBindingMode != nil && *sc.VolumeBindingMode == storagev1.VolumeBindingWaitForFirstConsumer {
+			continue
+		}
+		chosenPV, _, err = FindExistingOrCreateSimulatedBindableVolume(ctx, view, pvc, pvs, chosenPVs)
```
unmarshall: You could leverage component-helpers
https://github.com/kubernetes/component-helpers/blob/69b095369cfd7b7b7c13d9c55e38aa46300208c7/storage/volume/pv_helpers.go#L186C6-L186C24

common/volutil/volutil.go
```diff
@@ -0,0 +1,442 @@
...
+	pvs, err = viewutil.ListPersistentVolumes(ctx, view)
+	if err != nil {
+		return
+	}
+	for _, pv := range pvs {
+		ref := pv.Spec.ClaimRef
+		if ref == nil {
+			continue
+		}
+		if pv.Status.Phase == corev1.VolumeBound {
```
unmarshall: you can combine the two into a single condition since you are not logging or handling them differently.
```go
if ref == nil || pv.Status.Phase == corev1.VolumeBound {
  continue
}
```

unmarshall (commented):
4/n reviews

unmarshall (commented):
5/n reviews

planner/simulator/scaleout/singlenode/singlenode.go
```diff
@@ -0,0 +1,297 @@
...
+	schedulerLauncher plannerapi.SchedulerLauncher
+	storageMetaAccess plannerapi.StorageMetaAccess
+	nodeScorer        plannerapi.NodeScorer
+	traceDir          string
+	state             scaleout.RequestState
+	simulatorConfig   plannerapi.SimulatorConfig
+}
+
+// New creates a new plannerapi.ScaleOutSimulator that runs simulations for a single scaled node concurrently.
+func New(args plannerapi.SimulatorArgs) (plannerapi.ScaleOutSimulator, error) {
```
unmarshall: There is no need to return `error` from this function as it is always nil.

planner/simulator/scaleout/scaleout.go
```diff
@@ -0,0 +1,266 @@
...
+	"github.com/go-logr/logr"
+)
+
+var (
+	_ commontypes.Resettable = (*RequestState)(nil)
+)
+
+// RequestState holds the internal Request scoped state of a ScaleOutSimulator
+type RequestState struct {
+	viewAccess minkapi.ViewAccess
```
unmarshall: should `viewAccess` be considered as `RequestState`?

unmarshall (commented):
6/n review

api/common/types/types.go
elankath: Copied that.

common/objutil/objutil.go
```diff
@@ -359,6 +388,11 @@ func AsPtrMetaV1Time(t time.Time) *metav1.Time {
 	return mt
 }
 
+// AsStdTimeOrZero dereferences the given metav1.Time and returns the stdlib time.Time if not-nil or returns the zero value.
+func AsStdTimeOrZero(t *metav1.Time) time.Time {
```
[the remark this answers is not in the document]
elankath: Not sure why the lint did not catch this - it was being used before but usage was not needed later.

planner/simulator/scaleout/multinode/multinode.go
```diff
@@ -0,0 +1,123 @@
...
+		group := s.state.SimulationGroups[groupIndex]
+		log := log.WithValues("groupIndex", groupIndex, "groupName", group.Name())
+		grpCtx := logr.NewContext(ctx, log)
+		log.V(3).Info("Invoking group.Run")
+		s.runGroup(grpCtx, groupPassView, group)
+	}
+}
+
+func (s *simulatorSingleSim) runGroup(ctx context.Context, groupPassView minkapi.View, group plannerapi.ScaleOutSimGroup) (nextGroupPassView minkapi.View, err error) {
+	simResults, err := group.Run(ctx, func(ctx context.Context, name string) (minkapi.View, error) {
```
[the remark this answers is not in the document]
elankath: `multinode` has been deleted from this PR. it is meant for separate PR.

planner/simulator/scaleout/scaleout.go
```diff
@@ -0,0 +1,266 @@
...
+	// SimulationFactory is used to create `ScaleOutSimulation`s
+	SimulationFactory plannerapi.SimulationFactory
```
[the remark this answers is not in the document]
elankath: If you look at `ScaloutSimulator.Simulate` contract

```go
	// Simulate is the high level activity that runs [ScaleOutSimulation] created from given
	// [SimulationFactory] with the given planner [Request].
	Simulate(ctx context.Context, request *Request, simulationFactory SimulationFactory) (planResult <-chan ScaleOutPlanResult)
```

So yes, it is request state.

common/volutil/volutil.go
```diff
@@ -0,0 +1,487 @@
...
+		pv = pvObj.(*corev1.PersistentVolume)
+		log.V(4).Info("simulated PV already created for PVC", "pvName", pv.Name, "pvcName", pvc.Name, "pvcNamespace", pvc.Namespace)
+		return
+	}
+	if !apierrors.IsNotFound(err) {
+		return
+	}
+	pv = &corev1.PersistentVolume{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: simulatedVolumeNamePrefix + pvc.Namespace + "-" + pvc.Name,
```
takoverflow:
```suggestion
			Name: pvName,
```
This was already defined above

common/go.mod
```diff
@@ -12,19 +13,18 @@ require (
...
 	k8s.io/klog/v2 v2.130.1
+	k8s.io/kubernetes v1.35.2
+	k8s.io/utils v0.0.0-20251002143259-bc988d571ff4
 	sigs.k8s.io/yaml v1.6.0
 )
 
-replace (
-	github.com/gardener/scaling-advisor/api => ../api
-)
+replace github.com/gardener/scaling-advisor/api => ../api
```
takoverflow: There should be a replace directive for `minkapi` as well

common/volutil/volutil.go
```diff
@@ -0,0 +1,487 @@
...
+	log.V(3).Info("BindClaimsForImmediateMode succeeded", "numClaimAssignments", len(claimAssignments))
+	return
+}
+
+// ProvisionAndBindVolumesFoSelectedClaimsInWFFC performs provisioning of [corev1.PersistentVolume]'s for
+// [corev1.PersistentVolumeClaim]'s selected by the `kube-scheduler` for WaitForFirstConsumer volume binding.
+// It queries the given [minkapi.View] for PVC's that have been marked with [storagevolume.AnnSelectedNode] which
+// indicates that scheduler has triggered the PVC to be dynamically provisioned. It then creates a simulated virtual PV
+// that satisfies the PVC and also binds the same.
+func ProvisionAndBindVolumesFoSelectedClaimsInWFFC(ctx context.Context, view minkapi.View) (provisionPVs []*corev1.PersistentVolume, err error) {
```
takoverflow: Typo: `ProvisionAndBindVolumesForSelectedClaimsInWFFC` (for)

takoverflow (changes requested):

planner/simulation/scaleout/runstate.go
```diff
@@ -0,0 +1,359 @@
...
+			otherAssignments = append(otherAssignments, plannerapi.NodePodAssignment{
+				NodeResources: nodeResources,
+				ScheduledPods: scheduledPodInfos,
+			})
+		}
+	}
+	return otherAssignments, nil
+}
+
+func getUnscheduledPodsMap(ctx context.Context, v minkapi.View) (unscheduled map[commontypes.NamespacedName]plannerapi.PodResourceInfo, err error) {
```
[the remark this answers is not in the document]
elankath: This would not be the place to do this. We would need to do this when populating the review. However, I believe this should be taken up as optimization item for Multi Node Single Simulation PR

elankath (commented):
my comments were not visible earlier - they were all pending.

api/common/types/types.go
```diff
@@ -184,6 +220,31 @@ const (
...
 	// TraceLogPathCtxKey is the context key under which the path to the trace log file is stored.
-	TraceLogPathCtxKey ContextKey = "trace-log-path"
+	TraceLogPathCtxKey ContextKey = "trace-log"
 )
+
+// PriorityKey represents the key for ordering [NodePlacement]'s
+type PriorityKey struct {
+	// First is the first priority value. Higher values represent higher priority.
+	First int32
+	// Second is the 2nd priority value. Higher value represent higher priority. Secondary weight compared to First
```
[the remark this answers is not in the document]
elankath: Secondary weight implies exactly that - the weight is secondary relative to the First. Since type was moved to `api/common/types.go` and was made generic since neither NodePool nor NodePlacement are in this package.  I am changing the godoc to reflect this.

api/planner/scaleout_types.go
```diff
@@ -0,0 +1,229 @@
...
+	// Quota defines the resource quota for the node pool.
+	Quota corev1.ResourceList `json:"quota,omitempty"`
+	// Capacity defines the capacity for node resources that are available for the node's instance type.
+	Capacity corev1.ResourceList `json:"capacity"`
+	// KubeReserved defines the capacity for kube reserved resources.
+	// See https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#kube-reserved for additional information.
+	KubeReserved corev1.ResourceList `json:"kubeReservedCapacity,omitempty"`
+	// SystemReserved defines the capacity for system reserved resources.
+	// See https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#system-reserved for additional information.
+	// Please read https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#general-guidelines when deciding to
```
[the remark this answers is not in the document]
elankath: Removing redundant doc line

common/go.mod
elankath: No, common shouldn't depend on `minkapi` implementation Go module. Only on the `api` module. That dependency was spurious and removed it. `viewutil` was accidentally in `minkapi` impl module which caused the issue. Moved `viewutil` to `common` as it only depends on api.

planner/simulation/scaleout/runstate.go
```diff
@@ -0,0 +1,359 @@
...
+	numUnchangedTrackAttempts   int
+	numTrackAttempts            int
+	numReceivedEvents           int
+	numScheduledPods            int
+	runNum                      uint32
+	nodeCount                   atomic.Uint32
+}
+
+// FreshRunState returns a fresh RunState whose status is set to [plannerapi.ActivityStatusPending]
+func FreshRunState() RunState {
```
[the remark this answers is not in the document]
elankath: Anyways, I changed this to `MakeRunState` as `MakeXXX` is more common for factory functions that return a non-pointer instance.

takoverflow (approved):
The changes seem fine to me, thanks for bearing with the multiple sets of reviews. 

The ability to simulate volume provision is surely going to be a huge differentiator! Suggestions for further optimizations can be taken up later.

api/common/types/types.go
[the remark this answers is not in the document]
r4mek: @unmarshall 
This was the earlier discussion related to `PriorityKey`, `First` and `Second` which was referenced in #163

PR #154, opened 2026-04-01
ScaleOut plans using SimulatorStrategyMultiNodeSingleSim 

### LINKED ISSUE #85: [Feature] Simulator executing SimulatorStrategyMultiNodeSingleSim without Node Scoring
**How to categorize this issue?**

/area auto-scaling
/kind enhancement

- This simulator is used for the simulation strategy named `SimulationStrategySingleSimulationPerGroup=single-simulation-per-group`) and implements `ScaleOutSimulator.Simulate(ctx context.Context, resultCh chan<- ScalingPlanResult)`
- We create `SimGroupKey(s)`for all `NodePool` and `NodeTemplate` priority combinations.
- We create Simulations where each simulation is associated with multiple Node Pool(s), multiple Node Templates(s) and Availability Zones, where the Node Pool(s) and Node Template(s) belong to the same `SimGroupKey`.
- There is only Simulation per SimulationGroup.
- Each simulation is operated against a simulation view comprising of existing nodes and existing pods and unscheduled pods. When the simulation is run, it creates one or more simulated scaled node(s) per unique combination of NT + NP + AZ for all NTs + NPs + AZs associated with the SimulationGroup. 
- Runs a scheduler which performs pod-to-node assignments for all the simulated scaled nodes and existing node(s) for the leftover unscheduled pods.
-  We iterate through each SimulationGroup and concurrently run Simulation's within that group.
- If there are no leftover unscheduled pods, we produce the `service.ScalingPlanResult` from simulation state and emit then on results channel.
- If there are leftover unscheduled pods, then we move to the next `SimulationGroup` using the simulation view from the previous simulation.
- The simulator honors the provided Context when performing simulations

### Tasks

- [ ] Core implementation of simulator
- [ ] Unit Tests 
- [ ] Integration Tests of simulator against basic scenarios 
- [ ] Integration Tests  of simulator against standard scenarios

elankath:
* Introduced new simulator package planner/simulator/scaleout/multinode
* Enhanced planner/simulation/scaleout package to support new strategy
* changed to percentageOfNodesToScore as 100 for tests using new strategy

Fixes #85

**Release note**:

```feature user
ScaleOut plans using SimulatorStrategyMultiNodeSingleSim 
```


api/common/types/types.go
```diff
@@ -248,3 +248,6 @@ func CmpPriorityKeyDecreasing(a, b PriorityKey) int {
 	}
 	return cmp.Compare(b.Second, a.Second)
 }
+
+// ResourceIntValues represents aggregated resource quantities using int64 values.
+type ResourceIntValues map[corev1.ResourceName]int64
```
takoverflow: I don't see this type alias being used anywhere?

api/planner/constants.go
```diff
@@ -13,10 +13,10 @@ const (
 	// DefaultMaxParallelSimulations is the default maximum number of parallel simulations that can be run by the scaling advisor simulator.
 	DefaultMaxParallelSimulations = 1
 	// DefaultTrackPollInterval is the default polling interval for tracking pod scheduling in the view of the simulator.
-	DefaultTrackPollInterval = 40 * time.Millisecond
+	DefaultTrackPollInterval = 28 * time.Millisecond
 	//DefaultMaxUnchangedTrackAttempts is the default value for the maximum number of unchanged simulation track attempts after which
 	// a simulation run is considered as stabilized.
-	DefaultMaxUnchangedTrackAttempts = 65
+	DefaultMaxUnchangedTrackAttempts = 100
```
takoverflow, line 19: What was the reason for increasing this value to such a high number? Just curious about the reasoning.

api/planner/errors.go
```diff
@@ -24,7 +24,8 @@ var (
 	ErrCreateSimulation = errors.New("cannot create simulation")
 	// ErrRunSimulation is a sentinel error indicating that planner could not successfully run a specific scaling simulation
 	ErrRunSimulation = errors.New("cannot run simulation")
-
+	// ErrCreateSimNodes is a sentinel error indicating that the simulator failed to create simulation node(s).
```
takoverflow: Is there a reason why we add `is a sentinel error indicating` to each docstring? Can we not just simplify it to

```suggestion
	// ErrCreateSimNodes indicates that the simulator failed to create simulation node(s).
```

That is `s/is a sentinel error indicating/indicates/g`. I thought its inherently understood that these are all sentinel errors. WDYT? @elankath

common/logutil/logutil.go
```diff
@@ -82,6 +82,10 @@ func GetCleanLogFileName(fileName string) string {
 	return fileNameCleanRe.ReplaceAllString(fileName, "")
 }
 
+//func LogResourceList(log logr.Logger, resources corev1.ResourceList) {
```
takoverflow: Can this commented function be removed if its not going to be used

common/viewutil/viewutil.go
```diff
@@ -73,6 +74,9 @@ func ListPersistentVolumeClaims(ctx context.Context, view minkapi.View) ([]*core
 // At higher log verbosity, it also dumps all scheduling relevant objects into <tempDir>/<viewName>/ directory.
 func LogObjects(ctx context.Context, prefix string, view minkapi.View) error {
 	log := logr.FromContextOrDiscard(ctx)
+	if log.GetV() < 3 {
```
takoverflow, line 77: Just a question, not a review comment.

What is the default verbosity level we are targeting for the advisor? Is it `2` or `3`? Or just `V(1)`

planner/planner_test.go
```diff
@@ -39,6 +41,134 @@ func TestOnePoolUnitScaleOut(t *testing.T) {
...
+			{
+				NodePlacement: poolAPlacement,
+				Delta:         1,
+			},
+		},
+	}
+	testutil.ObtainAndAssertScaleOutPlan(t, planner, &testData, wantPlan)
+}
+
+func TestOnePoolFullFitPodScaleout(t *testing.T) {
```
takoverflow, line 68: Can these tests be suffixed with `_SingleNodeMultiSim` with the strategy explicitly passed as part of `testutil.Args`. I had to compare this with the next test in order to understand what the difference was.

The same comment is applicable for all these tests.

takoverflow: ~~Now that I think about it wouldn't it be better to combine these  in the form of a table driven test where you can just change the strategy and the expected delta. Since everything else remains the same? WDYT?~~

After today's discussion, I'm not sure you'd be in favour of TDT.

planner/scheduler/testdata/prefer-large-nodes-config.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: kubescheduler.config.k8s.io/v1
```
takoverflow: I don't see any difference between this and `scheduler/testdata/default-config.yaml`. Can you share the intent behind having this?

Consequently, I'm not sure what the value would be in having the tests for this config since its the same.

Also, should we be setting `percentageOfNodesToScore` to `100`?

planner/simulation/scaleout/scaleout.go
```diff
@@ -243,3 +246,66 @@ func (s *defaultSimulation) incRunNum() uint32 {
 	return s.args.RunCounter.Add(1)
 }
 
+// ComputeNodeTemplateCounts computes the number of nodes to scale-out for the given slice of
+// [plannerapi.ScaleOutNodeTemplate] and given slice of unscheduled [plannerapi.PodResourceInfo] in the scale-out simulation
+// It returns a slice of [NodeTemplateCount] which encapsulates the [plannerapi.ScaleOutNodeTemplate] and the scale-out count.
+//
+// At the moment, it uses simplistic logic, where
+//   - for [SimulatorStrategySingleNodeMultiSim], it just returns the a [NodeTemplateCount] using first [plannerapi.ScaleOutNodeTemplate] and count as `1`
```
takoverflow:
```suggestion
//   - for [SimulatorStrategySingleNodeMultiSim], it just returns a [NodeTemplateCount] using first [plannerapi.ScaleOutNodeTemplate] and count as `1`
```

planner/simulation/scaleout/scaleout.go
```diff
@@ -176,9 +178,10 @@ func (s *defaultSimulation) doWork(ctx context.Context, view minkapi.View) error
 		log.V(3).Info("FinalizeStaticBindingsForSelectedClaimsInWFFC performed work - reset RunState.numUnchangedTrackAttempts since ", "numBound", numBound)
 		s.state.numUnchangedTrackAttempts = 0
 	}
-	if s.args.Strategy.IsMultiNode() {
-		err = s.state.CreateSimulationNodes(s.args.StorageMetaAccess, s.args.NodeTemplates)
-	}
+	// TODO: needs discussion.
```
takoverflow: What's the discussion going to be about here? Are we not going to be creating simulation nodes, is it about the optimization where the nodetemplate predicates are checked against the unscheduled pods to see if it even makes sense to launch a simulation for this template?

planner/simulation/scaleout/scaleout.go
```diff
@@ -243,3 +246,66 @@ func (s *defaultSimulation) incRunNum() uint32 {
...
+	return plannerapi.NodeResourceInfo{
+		Name:         node.Name,
+		InstanceType: instanceType,
+		Capacity:     node.Status.Capacity,
+		Allocatable:  node.Status.Allocatable,
+	}
+}
+
+// NodeTemplateCount extends the [plannerapi.ScaleOutNodeTemplate] with a count representing the number of nodes to create
+// for this template
```
takoverflow: Might be worth mentioning in the docstring that the count is expected to be non-unit only for `MultiNodeSingleSim` strategy.

planner/simulator/scaleout/multinode/multinode.go
```diff
@@ -0,0 +1,147 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Package multinode provides implementation and helper routines of a ScaleOutSimulator that performs simulations that scale
+// multiple nodes at a time
```
takoverflow:
```suggestion
// multiple nodes for a single scale-out simulation.
```

planner/simulation/scaleout/scaleout.go
```diff
@@ -243,3 +246,66 @@ func (s *defaultSimulation) incRunNum() uint32 {
...
 }
 
+// ComputeNodeTemplateCounts computes the number of nodes to scale-out for the given slice of
+// [plannerapi.ScaleOutNodeTemplate] and given slice of unscheduled [plannerapi.PodResourceInfo] in the scale-out simulation
+// It returns a slice of [NodeTemplateCount] which encapsulates the [plannerapi.ScaleOutNodeTemplate] and the scale-out count.
+//
+// At the moment, it uses simplistic logic, where
+//   - for [SimulatorStrategySingleNodeMultiSim], it just returns the a [NodeTemplateCount] using first [plannerapi.ScaleOutNodeTemplate] and count as `1`
+//   - for [SimulatorStrategyMultiNodeSingleSim], it returns a [NodeTemplateCount] with count as the number of unscheduled pod(s) (multiplicative blowup of scale-out nodes)
+func ComputeNodeTemplateCounts(strategy commontypes.SimulatorStrategy, templates []plannerapi.ScaleOutNodeTemplate, podInfos []plannerapi.PodInfo) ([]NodeTemplateCount, error) {
```
takoverflow, line 252: Rather than having strategy based cases for computing the counts, wouldn't it be better to add a new function to the `ScaleOutSimulator` interface that can be implemented by each strategy to compute the `NodeTemplateCounts`. Similarly, IMO all strategy dependent functions should be declared in the `Simulator` interface rather than having functions to which the strategy is passed. WDYT?

This would make extending the simulator for a different strategy easier as well.

planner/simulator/scaleout/multinode/multinode.go
```diff
@@ -0,0 +1,147 @@
...
+type simulatorSingleSim struct {
+	viewAccess        minkapi.ViewAccess
+	schedulerLauncher plannerapi.SchedulerLauncher
+	storageMetaAccess plannerapi.StorageMetaAccess
+	state             *scaleout.SimulatorState
+	traceDir          string
+	simulatorConfig   plannerapi.SimulatorConfig
+}
+
+// New creates a new plannerapi.ScaleOutSimulator that runs simulations sequentially scaling multiple nodes from
```
takoverflow:
```suggestion
// New creates a new [plannerapi.ScaleOutSimulator] that runs simulations sequentially scaling multiple nodes from
```

planner/simulator/scaleout/multinode/multinode.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/gardener/scaling-advisor/api/minkapi"
+	plannerapi "github.com/gardener/scaling-advisor/api/planner"
+	"github.com/go-logr/logr"
+)
+
+var (
+	_ plannerapi.ScaleOutSimulator = (*simulatorSingleSim)(nil)
+)
+
+// simulatorSingleSim is a Simulator that implements ScaleOutSimulator for the SimulatorStrategyMultiNodeSingleSim.
```
takoverflow:
```suggestion
// simulatorSingleSim is a Simulator that implements [ScaleOutSimulator] for [SimulatorStrategyMultiNodeSingleSim].
```

planner/simulator/scaleout/multinode/multinode.go
```diff
@@ -0,0 +1,147 @@
...
+	} else {
+		nextGroupPassView = simResults[0].View // all simResults share the same View in this strategy
+	}
+	if s.state.Request.AdviceGenerationMode.IsIncremental() {
+		err = scaleout.SendPlanResultUsingSimResults(ctx, s.state.ResultCh, s.state.Request, s.state.SimRunCounter.Load(), simResults)
+	}
+	return
+}
+
+func (s *simulatorSingleSim) createAndGroupSimulations(ctx context.Context) ([]plannerapi.ScaleOutSimGroup, error) {
```
takoverflow: nit: Can this be moved before `runAllGroups()` so that all the functions are ordered in the sequence in which they are invoked

planner/simulator/scaleout/multinode/multinode.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	if s.state.Request.AdviceGenerationMode.IsAllAtOnce() {
+		err = scaleout.SendPlanResultUsingSimResults(ctx, s.state.ResultCh, s.state.Request, s.state.SimRunCounter.Load(), allSimResults)
+	}
+	return
+}
+
+func (s *simulatorSingleSim) runPassForGroup(ctx context.Context, group plannerapi.ScaleOutSimGroup, groupPassView minkapi.View) (simResults []plannerapi.ScaleOutSimResult, nextGroupPassView minkapi.View, err error) {
+	//var (
+	//	log = logr.FromContextOrDiscard(ctx)
```
takoverflow: Can this be removed

planner/simulator/scaleout/scaleout.go
```diff
@@ -139,32 +139,24 @@ func SendPlanError(planResultCh chan<- plannerapi.ScaleOutPlanResult, requestRef
 	}
 }
 
-// SendPlanResult creates a plannerapi.ScaleOutPlanResult from the given plannerapi.Request and plannerapi.SimulationGroupCycleResults
+// SendPlanResultUsingGroupCycleResults creates a plannerapi.ScaleOutPlanResult from the given plannerapi.Request and plannerapi.SimulationGroupCycleResults
```
takoverflow: Might be beneficial to add that this contains the node scores, I was a bit confused at first, why did we need two separate functions `SendPlanResultUsingGroupCycleResults` and `SendPlanResultUsingSimResults`.

Just by looking at the names and the docstrings it was not really clear the distinction between the two.

Also the function name doesn't really tell that its usage for sending plan results is highly coupled to the strategy being used.

```suggestion
// SendPlanResultUsingGroupCycleResults creates a [plannerapi.ScaleOutPlanResult] from the given [plannerapi.Request] and [plannerapi.SimulationGroupCycleResults] having winner node scores
```

Alternative since its again a strategy dependent way of sending the result i.e. `SendPlanResultUsingGroupCycleResults` is only used by singlenode and `SendPlanResultUsingSimResults` is only used by multinode; Why not add it to the `ScaleOutSimulator` interface as a `SendPlan()` function which needs to be implemented by each scale out simulator. That clearly shows what function is for which strategy.

Is the problem that one requires `[]plannerapi.ScaleOutSimGroupCycleResult` but the other needs `[]plannerapi.ScaleOutSimResult`?

planner/testutil/testutil.go
```diff
@@ -268,8 +268,8 @@ func CreateTestScalingPlanner(t *testing.T, args Args, traceDir string, verbosit
 	} else { // allow comfortable time for debugging
 		simulatorConfig = plannerapi.SimulatorConfig{
 			MaxParallelSimulations:    plannerapi.DefaultMaxParallelSimulations,
-			TrackPollInterval:         100 * plannerapi.DefaultTrackPollInterval,
-			MaxUnchangedTrackAttempts: 10 * plannerapi.DefaultMaxUnchangedTrackAttempts,
+			TrackPollInterval:         2 * plannerapi.DefaultTrackPollInterval,
```
takoverflow: In an earlier PR it was mentioned that with low values for these, debugging is harder. Are we sure that bringing them back to lower values won't cause any problems?

planner/simulator/scaleout/multinodesinglesim/simulator.go
```diff
@@ -0,0 +1,147 @@
...
+	})
+	if err != nil {
+		return
+	}
+	if len(simResults) == 0 {
+		nextGroupPassView = groupPassView
+		return
+	} else {
+		nextGroupPassView = simResults[0].View // all simResults share the same View in this strategy
+	}
```
takoverflow, line 138:
```suggestion
	if len(simResults) == 0 {
		nextGroupPassView = groupPassView
		return
	}
	// all simResults share the same View in this strategy
	nextGroupPassView = simResults[0].View 
```

takoverflow (commented):
Apologies for the late review. Have a couple of open points, PTAL.

Thanks for the changes!

api/common/types/types.go
elankath: was used transiently. removed now.

api/planner/constants.go
elankath: For the PV code. It takes time for the scheduler to do the binding on its side. Once the perf PR is raised this will matter less as we will exit early. Can discuss & implement other strategies here but not as part of this PR.

common/viewutil/viewutil.go
elankath: The default log verbosity will generally be `2`. But we will give an easy way to dynamically adjust it to `3` and above using the operator config map.

planner/simulator/scaleout/scaleout.go
elankath: I am moving  `SendPlanResultUsingGroupCycleResults`  and `SendPlanResultUsingSimResults` to the `singlenode` and `multinode` packages. When any other strategy needs to use the same method at that point we can move it back to the common `scaleout` package. Names seem OK if they are moved to the relevant packages.

I don't see it need to be added by the `ScaleoutSimulator` since this is clearly an internal implementation detail and not to be exposed by the consumers.

> Is the problem that one requires []plannerapi.ScaleOutSimGroupCycleResult but the other needs []plannerapi.ScaleOutSimResult?

No, why should it be ? Some simulators run multiple passes in a cycle and others do not.

planner/testutil/testutil.go
elankath: I am actually removing this code. Timeout needs to be adjusted by the guy debugging since for debugging some tests, greater timeouts are even *more* of a pain.

planner/simulator/scaleout/scaleout.go
takoverflow: > No, why should it be ? Some simulators run multiple passes in a cycle and others do not.

I meant if it was defined as an interface method. But since they've been moved to the corresponding packages, it's redundant.

planner/scheduler/testdata/prefer-large-nodes-config.yaml
elankath: Not needed. This was an approach `prefer-large-nodes-config.yaml` that was abandoned. Deleting

> should we be setting percentageOfNodesToScore to 100?

Yes, needed, otherwise weird results happen for large number of nodes. This is there in PR message too.

planner/simulation/scaleout/scaleout.go
elankath: No, the discussion was whether we should incrementally create nodes. Already had this discussion with madhav and the answer was that it was cheaper and more accurate to just create them upfront. Removin this TODO since it was already discussed.

planner/simulation/scaleout/scaleout.go
elankath: Yes, I will do this for next PR. technically this type is promoted to planner API in next PR.

planner/simulation/scaleout/scaleout.go
elankath: Yes, please wait for next PR - which has the concept of a `NodeEstimator` in the simulation args - which will take the strategy during construction.

planner/simulator/scaleout/multinodesinglesim/simulator.go
```diff
@@ -0,0 +1,176 @@
...
+		return
+	}
+	nextGroupPassView = simResults[0].View // all simResults share the same View in this strategy
+	if s.state.Request.AdviceGenerationMode.IsIncremental() {
+		err = sendPlanResultUsingSimResults(ctx, s.state.ResultCh, s.state.Request, s.state.SimRunCounter.Load(), simResults)
+	}
+	return
+}
+
+// sendPlanResultUsingSimResults constraints a [plannerapi.ScaleOutPlanResult] from the given slice of
```
thiyyakat, line 146: Is there a typo in the comment? constraints instead of constructs?
```suggestion
// sendPlanResultUsingSimResults constructs a [plannerapi.ScaleOutPlanResult] from the given slice of
```

planner/simulator/scaleout/singlenodemultisim/simulator.go
```diff
@@ -246,6 +246,34 @@ func (s *simulatorMultiSim) runPassForGroup(ctx context.Context, groupPassView m
 	return
 }
 
+// SendPlanResultUsingGroupCycleResults creates a plannerapi.ScaleOutPlanResult from the given plannerapi.Request and plannerapi.SimulationGroupCycleResults
```
thiyyakat, line 248:
```suggestion
// SendPlanResultUsingGroupCycleResults creates a [plannerapi.ScaleOutPlanResult] from the given [plannerapi.Request] and [plannerapi.SimulationGroupCycleResults]
```
nit: only if needed. We can also hover over the arguments too.

planner/simulation/scaleout/runstate.go
```diff
@@ -35,11 +34,10 @@ type RunState struct {
 	err                         error
 	ctx                         context.Context
 	view                        minkapi.View
-	scaleOutNodes               map[string]*corev1.Node                                   // map of node names to scale-out nodes
-	scaleOutPlacements          map[sacorev1alpha1.NodePlacement]int32                    // map of NodePlacement's to counts
-	unscheduledPods             map[commontypes.NamespacedName]plannerapi.PodResourceInfo // map of unscheduled Pod namespacedName to PodResourceInfo
-	scheduledPodNamesByNodeName map[string]sets.Set[commontypes.NamespacedName]           // map of node names to a set of scheduled pod names
-	leftoverUnscheduledPodNames sets.Set[commontypes.NamespacedName]                      // represents a set of pod names scheduled during simulation run
+	nodeAndPlacementByName      map[string]nodeAndPlacement                       // map of node names to scale-out nodes and their placement
```
thiyyakat, line 35: Could `Name` be changed to`NodeName`? If you feel the docstring is sufficient and the resultant variable name will be too long, please ignore this comment.

planner/scheduler/testdata/prefer-large-nodes-config.yaml
takoverflow: But the file `planner/scheduler/testdata/default-config.yaml` doesn't actually set it to `100`.

thiyyakat (commented):
Thanks! Just a few nits.

planner/simulator/scaleout/multinode/multinode.go
```diff
@@ -0,0 +1,176 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Package multinode provides implementation and helper routines of a ScaleOutSimulator that performs simulations that scale
+// multiple nodes for a single scale-out simulation
+package multinode
```
thiyyakat: As discussed offline, could you please rename the packages to `multinodesinglesim` and `singlenodemultisim`, the files to `simulator.go` and the structs to `defaultSimulator` ? Thanks!

PR #158, opened 2026-04-17
Fix Node Score Select() and corresponding tests for both strategies 

thiyyakat:

This PR fixes the flow in the scorer, and now computes the winning node score amongst all node scores rather than just the top node scores.

Fixes #

**Release note**:

```bugfix user
Fix scorer flow to consider all node scores in `Select()` rather than just the top valued scores.
```

### DIFF
--- a/api/planner/types.go
+++ b/api/planner/types.go
@@ -16,8 +16,8 @@ import (
 	sacorev1alpha1 "github.com/gardener/scaling-advisor/api/core/v1alpha1"
 	"github.com/gardener/scaling-advisor/api/minkapi"
 	"github.com/gardener/scaling-advisor/api/pricing"
-	"github.com/go-logr/logr"
 
+	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	nodev1 "k8s.io/api/node/v1"
 	schedulingv1 "k8s.io/api/scheduling/v1"
--- a/api/planner/types.go
+++ b/api/planner/types.go
@@ -16,6 +16,7 @@ import (
 	sacorev1alpha1 "github.com/gardener/scaling-advisor/api/core/v1alpha1"
 	"github.com/gardener/scaling-advisor/api/minkapi"
 	"github.com/gardener/scaling-advisor/api/pricing"
+	"github.com/go-logr/logr"
 
 	corev1 "k8s.io/api/core/v1"
 	nodev1 "k8s.io/api/node/v1"
@@ -289,7 +290,7 @@ type NodeScorer interface {
 	Compute(args NodeScorerArgs) (NodeScore, error)
 	// Select selects the winning NodeScore amongst the NodeScores of a given simulation pass and returns the pointer to the same.
 	// If there is no winning node score amongst the group, then it returns nil.
-	Select(groupNodeScores []NodeScore) (winningNodeScore *NodeScore, err error)
+	Select(log logr.Logger, groupNodeScores []NodeScore) (winningNodeScore *NodeScore, err error)
 }
 
 // NodeScorerArgs contains arguments for node scoring computation.
--- a/api/planner/types.go
+++ b/api/planner/types.go
@@ -17,7 +17,6 @@ import (
 	"github.com/gardener/scaling-advisor/api/minkapi"
 	"github.com/gardener/scaling-advisor/api/pricing"
 
-	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	nodev1 "k8s.io/api/node/v1"
 	schedulingv1 "k8s.io/api/scheduling/v1"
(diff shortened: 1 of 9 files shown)


planner/scorer/scorer.go
```diff
@@ -73,20 +74,37 @@ func (l LeastCost) Compute(args plannerapi.NodeScorerArgs) (score plannerapi.Nod
...
+func (l LeastCost) Select(log logr.Logger, nodeScores []plannerapi.NodeScore) (*plannerapi.NodeScore, error) {
 	if len(nodeScores) == 0 {
 		return nil, plannerapi.ErrNoWinningNodeScore
 	}
 	if len(nodeScores) == 1 {
+		log.V(4).Info("Single node score, selected directly", "templateName", nodeScores[0].Placement.TemplateName, "instanceType", nodeScores[0].Placement.InstanceType)
 		return &nodeScores[0], nil
 	}
+
+	// first pass: collect indices of node scores with the highest value
```
takoverflow: Why don't we just sort the nodeScores by `Value` (applicable regardless of the strategy) before calling `Select()`

Thus ridding ourselves of the need of doing two passes. For multiple node scores of the same value, we just need to check from the start till the point the scores reduced/changed to pick the most suitable. (Whatever data is being logged doesn't need to change as well, and can be computed during this single pass)

Rather than constructing this list of indices which keeps on getting re-written each time a higher score is seen.

For LeastCost we iterate in order, whereas for LeastWaste we iterate in reverse order. WDYT?

thiyyakat: I will replace the first pass with the sort. 

> Why don't we just sort the nodeScores by Value (applicable regardless of the strategy) before calling Select()

While this may improve code reusability, I feel doing the sort in Select makes the code in Select easier to understand since the logic depends on the ordering of the nodes. It will seem extremely arbitrary otherwise.

takoverflow: Makes sense.

api/planner/types.go
```diff
@@ -289,7 +290,7 @@ type NodeScorer interface {
 	Compute(args NodeScorerArgs) (NodeScore, error)
 	// Select selects the winning NodeScore amongst the NodeScores of a given simulation pass and returns the pointer to the same.
 	// If there is no winning node score amongst the group, then it returns nil.
-	Select(groupNodeScores []NodeScore) (winningNodeScore *NodeScore, err error)
+	Select(log logr.Logger, groupNodeScores []NodeScore) (winningNodeScore *NodeScore, err error)
```
thiyyakat: Team review: pass context.Context instead

api/planner/types.go
```diff
@@ -289,7 +290,7 @@ type NodeScorer interface {
 	Compute(args NodeScorerArgs) (NodeScore, error)
```
thiyyakat: Pass context here

planner/scorer/scorer.go
```diff
@@ -73,35 +76,47 @@ func (l LeastCost) Compute(args plannerapi.NodeScorerArgs) (score plannerapi.Nod
...
-func (l LeastCost) Select(nodeScores []plannerapi.NodeScore) (*plannerapi.NodeScore, error) {
+func (l LeastCost) Select(log logr.Logger, nodeScores []plannerapi.NodeScore) (*plannerapi.NodeScore, error) {
 	if len(nodeScores) == 0 {
 		return nil, plannerapi.ErrNoWinningNodeScore
 	}
 	if len(nodeScores) == 1 {
+		log.V(4).Info("Single node score, selected directly", "templateName", nodeScores[0].Placement.TemplateName, "instanceType", nodeScores[0].Placement.InstanceType)
 		return &nodeScores[0], nil
 	}
+	slices.SortStableFunc(nodeScores, func(a, b plannerapi.NodeScore) int {
```
thiyyakat, line 103: Sort from max to min instead. No need to maintain consistency across Select() implementations.

planner/scorer/scorer.go
```diff
@@ -73,35 +76,47 @@ func (l LeastCost) Compute(args plannerapi.NodeScorerArgs) (score plannerapi.Nod
...
+	for i = len(nodeScores) - 1; i >= 0; i-- {
+		if nodeScores[i].Value != nodeScores[len(nodeScores)-1].Value {
+			break
+		}
+		candidate := nodeScores[i]
 		weights, err := l.resourceWeigher.GetWeights(candidate.Placement.InstanceType)
 		if err != nil {
 			return nil, err
 		}
 		normalizedAlloc := getNormalizedResourceUnits(candidate.ScaledNodeResource.Allocatable, weights)
```
thiyyakat: Rename to getNormalizedAggregate()

planner/scorer/scorer.go
```diff
@@ -73,35 +76,47 @@ func (l LeastCost) Compute(args plannerapi.NodeScorerArgs) (score plannerapi.Nod
...
 		if maxNormalizedAlloc == normalizedAlloc {
-			winnerIndices = append(winnerIndices, index)
+			winnerIndices = append(winnerIndices, i)
```
thiyyakat: Replace with a second sort based on allocatable.

planner/scorer/scorer.go
```diff
@@ -73,35 +76,47 @@ func (l LeastCost) Compute(args plannerapi.NodeScorerArgs) (score plannerapi.Nod
...
 			winnerIndices = winnerIndices[:0]
-			winnerIndices = append(winnerIndices, index)
+			winnerIndices = append(winnerIndices, i)
 			maxNormalizedAlloc = normalizedAlloc
 		}
 	}
+	log.V(5).Info("Tie-break by allocatable resources", "numTopValueScores", len(nodeScores)-i-1, "numMaxAllocScores", len(winnerIndices), "maxNormalizedAlloc", maxNormalizedAlloc)
 	//pick one winner at random from winnerIndices
 	randIndex := rand.IntN(len(winnerIndices)) // #nosec G404 -- cryptographic randomness not required here. It randomly picks one of the node scores with the same least price.
+	log.V(4).Info("Winner node score", "scoreValue", nodeScores[winnerIndices[randIndex]].Value, "templateName", nodeScores[winnerIndices[randIndex]].Placement.TemplateName, "instanceType", nodeScores[winnerIndices[randIndex]].Placement.InstanceType)
```
thiyyakat: Change this to log level 3

thiyyakat (commented):
Team review

elankath (approved):
/lgtm

PR #164, opened 2026-05-16
Update genscenario (CA priority expander, Volume data)

### LINKED ISSUE #49: [Feature] Tool to generate scaling scenarios
**How to categorize this issue?**

/area auto-scaling|
/kind enhancement

We need the ability to generate Scaling Scenario's from real-world cluster. For now, we propose suggesting only gardener clusters, but tool must be generic enough to support scaling scenario generation for any cluster manager (cluster-api, etc), not just gardener.

#### CLI  Tool `scadctl genscenario`

NOTE: `scadctl genscenario` is an auxiliary command for the scaling-advisor ctl client `scadctl` that generates scaling advisor scenarios from gardener shoot clusters.

##### Code Placement
1.  Introduce cobra command `genscenario` inside `tools/cmd/scenario/genscenario.go`
 2. Register cobra command into root.
 3. gardener cluster manager scenario generation code should go into `tools/cmd/scenario/gardener`.  scenario generation for other cluster managers in the future (`cluster-api`) can go into subpackages like: `client/cmd/scenario/<cluster-manager>`. Ex: `client/cmd/scenario/clusterapi` 

##### Command Overview
1. `scadctl genscenario (-m|--cluster-manager) <cluster-manager> <cluster-manager-specific-options>`
     1. For gardener clusters: `scadctl genscenario -m gardener  -l <landscape> -p <project> -t <shoot-name> -d <scenario-dir>`.  # capture cluster state and generates scenarios from live cluster

##### `scadctl genscenario -m gardener  -l <landscape> -p <project> -t <shoot-name> -d <scenario-dir>`

1. generates viewer kubeconfig for shoot data planea and control plane into OS `tmp/full-qualified-clustername` directory if not already generated. 
	1. Delegate to gardenctl CLI tool for this or use HTTP request to generate kubeconfigs. Do not introduce gardener code dependency.
	2. Example of `full-qualified-clustername -> dev:i034796:aw`
2.  Connect to cluster and create a `svcapi.ClusterSnapshot`.  (See Scaling Advisor Service API)
3. Connect to control cluster, get the `Worker` object and then create the `sacorev1alpha1.aClusterScalingConstraints` (See Scaling Advisor Common Core API)
4. (optional) It should obfuscate the node names and pod names and pool names. Use generated human names component. 
5. Create the `scenario-dir` if it does not exist.
6. Persist the the `sacorev1alpha1.ClusterScalingConstraints` into `scenario-dir/<full-qualified-cluster-name>/cluster-scaling-constraints.json`. 
7. It should now create cluster snapshot variants:
	1. `scenario-dir/clusterName/snapshot-ddMMYY-latest-incremental.json`
		1. From the captured cluster snapshot. remove the latest created node , unassign the pods from this node.  Rewrite the cluster snapshot
	2. `scenario-dir/clusterName/snapshot-ddMMYY-latest-5.json`
		1. From the captured cluster snapshot. remove the latest 5 created node , unassign the pods from this node. Rewrite the cluster snapshot
	3. (Redo the above for 5-10-20 nodes)
8. The `scenario-dir` can be checked into `github.tools.sap/kubernetes/gardener-scaling-tools/scenarios` to allow other gardener members to reply and reproduce issues in scenarios via the `scalebench` tool described later. (If scenarios are too large, we can either compress and. check them in or we can attach them as github release artifacts instead of direct source-checkin)

```
/scenario-dir
   /<full-qualified-cluster-name>
	   /cluster-scaling-constraints.json
	   /snapshot-ddMMYY-latest-incremental.json
	   /snapshot-ddMMYY-latest-5.json
	   ...
```

**Why is this needed**:

We need to test scaling advisor (and other scalers) against actual scaling scenarios. For the gardener cluster-manager, we need the ability to adapt the gardener shoot worker object and the state of the shoot cluster into a `svcapi.ScalingAdviceRequest` that the scaling-advisor can be invoked with.

takoverflow:
1. The PR fixes the bug where a shoot using `NamespacedCloudProfile` wouldn't get its list of supported instances and cause the `genscenario` command to fail.

2. Support for defining `NodePool` priorities by leveraging `cluster-autoscaler-priority-expander` ConfigMap (when present) has been added and the same is exported when generating the scenario as well.

3. Additional bug fixes include cleanup of pods before persisting into snapshot, fixing incorrect TSCs so that the `apiserver` doesn't throw validation errors when deploying the pods.

4. `genscenario` command for gardener shoots now also captures the relevant volume related data for the cluster snapshot i.e. PVs, PVCs and StorageClasses.

Fixes #49

Would recommend looking at the PR commit-wise for easier comprehension of the changes.

**Release note**:

```feature user

```

### DIFF
--- a/api/planner/types.go
+++ b/api/planner/types.go
@@ -126,6 +126,9 @@ type ClusterSnapshot struct {
 	Pods []PodInfo `json:"pods,omitempty"`
 	// Nodes are the nodes that are present in the cluster.
 	Nodes []NodeInfo `json:"nodes,omitempty"`
+	// PodOwners contain the metadata and the replica count for each kind of pod ownerReference.
+	// Currently includes ReplicaSet, StatefulSet and Jobs.
+	PodOwners []PodOwnerInfo `json:"podOwners,omitempty"`
 	// PVs are the information about PersistentVolumes in the cluster. Should not contain deleted PVs.
 	// Should only contain *bound* PVs ie those with populated claimRef.
 	PVs []PVInfo `json:"pvs,omitempty"`
@@ -259,6 +262,20 @@ func (n *NodeInfo) GetNodePlacement() (placement sacorev1alpha1.NodePlacement, e
 		AvailabilityZone: n.Labels[corev1.LabelTopologyZone],
 	}
 	return
+}
+
+// PodOwnerInfo encapsulates the different owners for a pod object with the minimal
+// metadata necessary for reconstruction of the owner object. This includes owners such
+// as 'ReplicaSet', 'StatefulSet', 'Job' etc.
+// This data is currently used to allow for the benchmarking harness to reconstruct the
+// pod owners to allow for cluster-autoscaler to function properly for scale-in activities.
+type PodOwnerInfo struct {
+	Name            string                `json:"name"`
+	Namespace       string                `json:"namespace"`
+	Kind            string                `json:"kind"`
+	Selector        *metav1.LabelSelector `json:"selector,omitzero"`
+	TargetReplicas  *int32                `json:targetReplicas,omitzero"`
+	CurrentReplicas int32                 `json:currentReplicas,omitzero"`
 }
 
 // PVCInfo encapsulates the minimal set of scheduling relevant information about the k8s PersistentVolumeClaim.
--- a/api/planner/types.go
+++ b/api/planner/types.go
@@ -267,8 +267,8 @@ func (n *NodeInfo) GetNodePlacement() (placement sacorev1alpha1.NodePlacement, e
 // PodOwnerInfo encapsulates the different owners for a pod object with the minimal
 // metadata necessary for reconstruction of the owner object. This includes owners such
 // as 'ReplicaSet', 'StatefulSet', 'Job' etc.
-// This data is currently used to allow for the benchmarking harness to reconstruct the
-// pod owners to allow for cluster-autoscaler to function properly for scale-in activities.
+// This data is currently used by the benchmarking harness to reconstruct the
+// pod owners, allowing cluster-autoscaler to perform scale-in correctly.
 type PodOwnerInfo struct {
 	Selector        *metav1.LabelSelector `json:"selector,omitzero"`
 	TargetReplicas  *int32                `json:"targetReplicas,omitzero"`
--- a/api/planner/types.go
+++ b/api/planner/types.go
@@ -270,12 +270,12 @@ func (n *NodeInfo) GetNodePlacement() (placement sacorev1alpha1.NodePlacement, e
 // This data is currently used to allow for the benchmarking harness to reconstruct the
 // pod owners to allow for cluster-autoscaler to function properly for scale-in activities.
 type PodOwnerInfo struct {
+	Selector        *metav1.LabelSelector `json:"selector,omitzero"`
+	TargetReplicas  *int32                `json:"targetReplicas,omitzero"`
 	Name            string                `json:"name"`
 	Namespace       string                `json:"namespace"`
 	Kind            string                `json:"kind"`
-	Selector        *metav1.LabelSelector `json:"selector,omitzero"`
-	TargetReplicas  *int32                `json:targetReplicas,omitzero"`
-	CurrentReplicas int32                 `json:currentReplicas,omitzero"`
+	CurrentReplicas int32                 `json:"currentReplicas,omitzero"`
 }
 
 // PVCInfo encapsulates the minimal set of scheduling relevant information about the k8s PersistentVolumeClaim.
--- a/api/planner/types.go
+++ b/api/planner/types.go
@@ -126,9 +126,6 @@ type ClusterSnapshot struct {
 	Pods []PodInfo `json:"pods,omitempty"`
 	// Nodes are the nodes that are present in the cluster.
 	Nodes []NodeInfo `json:"nodes,omitempty"`
-	// PodOwners contain the metadata and the replica count for each kind of pod ownerReference.
-	// Currently includes ReplicaSet, StatefulSet and Jobs.
-	PodOwners []PodOwnerInfo `json:"podOwners,omitempty"`
 	// PVs are the information about PersistentVolumes in the cluster. Should not contain deleted PVs.
 	// Should only contain *bound* PVs ie those with populated claimRef.
 	PVs []PVInfo `json:"pvs,omitempty"`
@@ -262,20 +259,6 @@ func (n *NodeInfo) GetNodePlacement() (placement sacorev1alpha1.NodePlacement, e
 		AvailabilityZone: n.Labels[corev1.LabelTopologyZone],
 	}
 	return
-}
-
-// PodOwnerInfo encapsulates the different owners for a pod object with the minimal
-// metadata necessary for reconstruction of the owner object. This includes owners such
-// as 'ReplicaSet', 'StatefulSet', 'Job' etc.
-// This data is currently used by the benchmarking harness to reconstruct the
-// pod owners, allowing cluster-autoscaler to perform scale-in correctly.
-type PodOwnerInfo struct {
-	Selector        *metav1.LabelSelector `json:"selector,omitzero"`
-	TargetReplicas  *int32                `json:"targetReplicas,omitzero"`
-	Name            string                `json:"name"`
-	Namespace       string                `json:"namespace"`
-	Kind            string                `json:"kind"`
-	CurrentReplicas int32                 `json:"currentReplicas,omitzero"`
 }
 
 // PVCInfo encapsulates the minimal set of scheduling relevant information about the k8s PersistentVolumeClaim.
--- a/common/volutil/volutil.go
+++ b/common/volutil/volutil.go
@@ -40,6 +40,8 @@ func AsPVInfo(pv corev1.PersistentVolume) plannerapi.PVInfo {
 		StorageClassName: pv.Spec.StorageClassName,
 		Phase:            pv.Status.Phase,
 	}
+	pvi.ResourceVersion = ""
+	pvi.ManagedFields = nil
 	if pv.Spec.ClaimRef != nil {
 		pvi.ClaimRef.Namespace = pv.Spec.ClaimRef.Namespace
 
@@ -77,11 +79,14 @@ func AsPV(p plannerapi.PVInfo) *corev1.PersistentVolume {
 
 // AsPVCInfo converts the given k8s corev1.PersistentVolumeClaim object to a lean planner PVCInfo.
 func AsPVCInfo(pvc corev1.PersistentVolumeClaim) plannerapi.PVCInfo {
-	return plannerapi.PVCInfo{
+	pvcInfo := plannerapi.PVCInfo{
 		ObjectMeta:                pvc.ObjectMeta,
 		PersistentVolumeClaimSpec: pvc.Spec,
 		Phase:                     pvc.Status.Phase,
 	}
+	pvcInfo.ResourceVersion = ""
+	pvcInfo.ManagedFields = nil
+	return pvcInfo
 }
 
 // AsPVC converts the given plannerapi PVCInfo object to a corev1.PersistentVolumeClaim
--- a/tools/cmd/scadctl/cmd/genscenario/gardener.go
+++ b/tools/cmd/scadctl/cmd/genscenario/gardener.go
@@ -29,6 +29,7 @@ import (
 	"github.com/gardener/scaling-advisor/common/nodeutil"
 	"github.com/gardener/scaling-advisor/common/objutil"
 	"github.com/gardener/scaling-advisor/common/podutil"
+	"github.com/gardener/scaling-advisor/common/volutil"
 	"github.com/spf13/cobra"
 	"go.yaml.in/yaml/v2"
 	appsv1 "k8s.io/api/apps/v1"
@@ -74,6 +75,12 @@ type ShootAccess interface {
 	ListPods(ctx context.Context, criteria minkapi.MatchCriteria, excludeSystemComponents bool) ([]corev1.Pod, error)
 	// ListPodOwners fetches the pod owners on a shoot cluster having more than 0 desired replicas.
 	ListPodOwners(ctx context.Context, excludeSystemComponents bool) ([]planner.PodOwnerInfo, error)
+	// ListPVs fetches the persistent volumes on a shoot cluster.
+	ListPVs(ctx context.Context) ([]planner.PVInfo, error)
+	// ListPVCs fetches the persistent volume claims on a shoot cluster.
+	ListPVCs(ctx context.Context, excludeSystemComponents bool) ([]planner.PVCInfo, error)
+	// ListStorageClasses fetches the storage classes on a shoot cluster.
+	ListStorageClasses(ctx context.Context) ([]storagev1.StorageClass, error)
 	// ListPriorityClasses fetches all the priority classes present on a shoot cluster.
 	ListPriorityClasses(ctx context.Context, excludeSystemComponents bool) ([]schedulingv1.PriorityClass, error)
 	// ListRuntimeClasses fetches all the runtime classes present on a shoot cluster.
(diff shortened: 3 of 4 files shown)


tools/cmd/scadctl/cmd/genscenario/gardener.go
```diff
@@ -612,6 +767,43 @@ func (a *access) createNodePools() (nodePools []sacorev1alpha1.NodePool) {
...
+	var config map[int][]string
+	if err := yaml.Unmarshal([]byte(priorityStr), &config); err != nil {
+		return nil, fmt.Errorf("can't parse YAML with priorities in the configmap: %v", err)
+	}
+	priorities := make(caPrioritiesMap)
+	for prio, reList := range config {
+		for _, re := range reList {
+			regexp, err := regexp.Compile(re)
+			if err != nil {
+				return nil, fmt.Errorf("Can't compile regexp rule for priority %d and rule %s: %v", prio, re, err)
```
thiyyakat: nit: error string starting with an uppercase letter.

tools/cmd/scadctl/cmd/genscenario/gardener.go
```diff
@@ -612,6 +767,43 @@ func (a *access) createNodePools() (nodePools []sacorev1alpha1.NodePool) {
...
 
+func constructPrioritiesFromConfigMap(caPriorityCM corev1.ConfigMap) (caPrioritiesMap, error) {
+	priorityStr, found := caPriorityCM.Data["priorities"]
+	if !found {
+		return nil, fmt.Errorf("incorrect configmap for priority expander")
+	}
+
+	var config map[int][]string
+	if err := yaml.Unmarshal([]byte(priorityStr), &config); err != nil {
+		return nil, fmt.Errorf("can't parse YAML with priorities in the configmap: %v", err)
```
thiyyakat: Could you reword the error message for easier comprehension, please? Maybe `"failed to unmarshal priorities configmap: %v"`?

thiyyakat (commented):
Thanks for the PR. Just nits.

api/planner/types.go
```diff
@@ -261,6 +264,20 @@ func (n *NodeInfo) GetNodePlacement() (placement sacorev1alpha1.NodePlacement, e
 	return
 }
 
+// PodOwnerInfo encapsulates the different owners for a pod object with the minimal
+// metadata necessary for reconstruction of the owner object. This includes owners such
+// as 'ReplicaSet', 'StatefulSet', 'Job' etc.
+// This data is currently used to allow for the benchmarking harness to reconstruct the
+// pod owners to allow for cluster-autoscaler to function properly for scale-in activities.
```
thiyyakat: WDYT?

```suggestion
// This data is currently used by the benchmarking harness to reconstruct the
// pod owners, allowing cluster-autoscaler to perform scale-in correctly.
```

tools/cmd/scadctl/cmd/genscenario/gardener.go
```diff
@@ -323,6 +358,23 @@ func getClient(kubeConfigPath string, scheme *runtime.Scheme) (client.Client, er
...
 }
 
+func getCAPriorityExpanderConfigMap(ctx context.Context, shootClient client.Client) (*corev1.ConfigMap, error) {
+	var caPriorityCM corev1.ConfigMap
+	err := shootClient.Get(ctx, client.ObjectKey{
+		Name:      "cluster-autoscaler-priority-expander",
+		Namespace: "kube-system",
+	}, &caPriorityCM)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, fmt.Errorf("error getting the priority expander configmap: %v", err)
```
elankath, line 364: `return nil, fmt.Errorf("failed to get priority expander configmap: %w", err)`

tools/cmd/scadctl/cmd/genscenario/gardener.go
```diff
@@ -347,12 +399,16 @@ func (a *access) ListNodes(ctx context.Context, criteria minkapi.MatchCriteria)
 	return
 }
 
-func (a *access) ListPods(ctx context.Context, criteria minkapi.MatchCriteria, excludeKubeSystemPods bool) (pods []corev1.Pod, err error) {
+func (a *access) ListPods(ctx context.Context, criteria minkapi.MatchCriteria, excludeSystemComponents bool) (pods []corev1.Pod, err error) {
```
elankath, line 398: We should extene `MatchCriteria` with a `FieldSelector labels.Selector` instead of this . Infact it is mentioned as `minkapi` `TODO` somehwere.

tools/cmd/scadctl/cmd/genscenario/gardener.go
```diff
@@ -539,24 +695,86 @@ func removeNodesFromSnapshot(snap planner.ClusterSnapshot, count int) planner.Cl
 	return newSnap
 }
 
+func addNodesToSnapshot(
```
elankath, line 615: Be aware that this logic will change when aaron merges api changes.

tools/cmd/scadctl/cmd/genscenario/gardener.go
takoverflow: Agreed, I'll update `MatchCriteria`

tools/cmd/scadctl/cmd/genscenario/gardener.go
```diff
@@ -720,6 +1043,11 @@ func sanitizeNode(node *corev1.Node) {
...
 }
 
+func sanitizeStorageClass(sc *storagev1.StorageClass) {
+	sc.ManagedFields = nil
+	sc.ResourceVersion = ""
+}
+
 func sanitizePriorityClass(pc *schedulingv1.PriorityClass) {
 	pc.ManagedFields = nil
 	pc.ResourceVersion = ""
```
elankath, line 970: Please use a single function to set managedfields & resource version to empty. Introduce utiltity function `ResetObjMetadata` in `objutil` and use it everywhere you need. No need for repetition.

```go
func ResetObjMetadata(obj metav1.Object) {
	obj.SetManagedFields(nil)
	obj.SetResourceVersion("")
}
```

api/planner/types.go
```diff
@@ -261,6 +264,20 @@ func (n *NodeInfo) GetNodePlacement() (placement sacorev1alpha1.NodePlacement, e
 	return
 }
 
+// PodOwnerInfo encapsulates the different owners for a pod object with the minimal
+// metadata necessary for reconstruction of the owner object. This includes owners such
+// as 'ReplicaSet', 'StatefulSet', 'Job' etc.
+// This data is currently used by the benchmarking harness to reconstruct the
+// pod owners, allowing cluster-autoscaler to perform scale-in correctly.
+type PodOwnerInfo struct {
```
unmarshall: Why is it required to have a separate `PodOwnerInfo` and why we cannot include this as part of the `PodInfo`?

api/planner/types.go
```diff
@@ -261,6 +264,20 @@ func (n *NodeInfo) GetNodePlacement() (placement sacorev1alpha1.NodePlacement, e
...
+	Selector        *metav1.LabelSelector `json:"selector,omitzero"`
+	TargetReplicas  *int32                `json:"targetReplicas,omitzero"`
```
unmarshall: What is `TargetReplicas`?

api/planner/types.go
```diff
@@ -261,6 +264,20 @@ func (n *NodeInfo) GetNodePlacement() (placement sacorev1alpha1.NodePlacement, e
 	return
 }
 
+// PodOwnerInfo encapsulates the different owners for a pod object with the minimal
+// metadata necessary for reconstruction of the owner object. This includes owners such
+// as 'ReplicaSet', 'StatefulSet', 'Job' etc.
+// This data is currently used by the benchmarking harness to reconstruct the
+// pod owners, allowing cluster-autoscaler to perform scale-in correctly.
+type PodOwnerInfo struct {
+	Selector        *metav1.LabelSelector `json:"selector,omitzero"`
```
unmarshall: What is the selector meant to select?
When you already have the name, namespace and kind then why is a selector additionally required?

api/planner/types.go
```diff
@@ -261,6 +264,20 @@ func (n *NodeInfo) GetNodePlacement() (placement sacorev1alpha1.NodePlacement, e
...
+	TargetReplicas  *int32                `json:"targetReplicas,omitzero"`
+	Name            string                `json:"name"`
+	Namespace       string                `json:"namespace"`
+	Kind            string                `json:"kind"`
+	CurrentReplicas int32                 `json:"currentReplicas,omitzero"`
```
unmarshall: why is this required?

common/volutil/volutil.go
```diff
@@ -40,6 +40,8 @@ func AsPVInfo(pv corev1.PersistentVolume) plannerapi.PVInfo {
 		StorageClassName: pv.Spec.StorageClassName,
 		Phase:            pv.Status.Phase,
 	}
+	pvi.ResourceVersion = ""
```
unmarshall, line 43: Why is there a need to set the ResourceVersion as empty?

common/volutil/volutil.go
```diff
@@ -40,6 +40,8 @@ func AsPVInfo(pv corev1.PersistentVolume) plannerapi.PVInfo {
...
+	pvi.ManagedFields = nil
```
unmarshall, line 44: @elankath perhaps we do not need the entirety of the ObjectMeta in `PVInfo` and `PVCInfo`. Can we just have what is required instead? I find if quite unnecessary to first have `PVInfo` on the pretext that we do not need all fields but continue to keep the entire ObjectMeta and then have such code to assign nil/empty values to fields.

tools/cmd/scadctl/cmd/genscenario/gardener.go
```diff
@@ -64,9 +72,17 @@ type ShootAccess interface {
 	// ListNodes fetches the nodes on a shoot cluster matching the given criteria.
 	ListNodes(ctx context.Context, criteria minkapi.MatchCriteria) ([]corev1.Node, error)
 	// ListPods fetches the pods on a shoot cluster matching the given criteria.
-	ListPods(ctx context.Context, criteria minkapi.MatchCriteria, excludeKubeSystemPods bool) ([]corev1.Pod, error)
+	ListPods(ctx context.Context, criteria minkapi.MatchCriteria, excludeSystemComponents bool) ([]corev1.Pod, error)
```
unmarshall, line 73: Are you subsuming `kube-system` pods into `system-components`?

tools/cmd/scadctl/cmd/genscenario/gardener.go
```diff
@@ -64,9 +72,17 @@ type ShootAccess interface {
...
+	// ListPodOwners fetches the pod owners on a shoot cluster having more than 0 desired replicas.
+	ListPodOwners(ctx context.Context, excludeSystemComponents bool) ([]planner.PodOwnerInfo, error)
```
unmarshall: why is this required? Neither in the issue nor in the PR there is any mention of the use case. Can you please explicitly clarify as this serves as clear and proper documentation for any future reference

unmarshall (changes requested):

api/planner/types.go
```diff
@@ -126,6 +126,9 @@ type ClusterSnapshot struct {
 	Pods []PodInfo `json:"pods,omitempty"`
 	// Nodes are the nodes that are present in the cluster.
 	Nodes []NodeInfo `json:"nodes,omitempty"`
+	// PodOwners contain the metadata and the replica count for each kind of pod ownerReference.
+	// Currently includes ReplicaSet, StatefulSet and Jobs.
+	PodOwners []PodOwnerInfo `json:"podOwners,omitempty"`
```
aaronfern: Daemonset is also a valid pod owner. Is there a reason this was not considered?

tools/cmd/scadctl/cmd/genscenario/gardener.go
```diff
@@ -323,6 +349,23 @@ func getClient(kubeConfigPath string, scheme *runtime.Scheme) (client.Client, er
...
+		Namespace: "kube-system",
+	}, &caPriorityCM)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, fmt.Errorf("error getting the priority expander configmap: %v", err)
+	} else if apierrors.IsNotFound(err) {
+		return nil, nil
+	}
+	caPriorityCM.Annotations = nil
+	caPriorityCM.ManagedFields = nil
+	caPriorityCM.ResourceVersion = ""
```
aaronfern, line 370: Why not remove fields like `creationTimestamp` and `uid`?

tools/cmd/scadctl/cmd/genscenario/gardener.go
```diff
@@ -900,6 +901,38 @@ func sanitizePod(pod *corev1.Pod) {
...
 	}
+
+	// This is needed to clean up pods that cannot be deployed due to failures:
+	// metadata.annotations[container.apparmor.security.beta.kubernetes.io/install-cni]:
+	// Invalid value: "install-cni": container not found
+	// Required as PodInfo doesn't track all the containers in the pod object. Rather
+	// all the container resources are combined to form a single dummy aggregated container.
+	maps.DeleteFunc(pod.Annotations, func(k, _ string) bool {
+		return strings.HasSuffix(k, "install-cni")
+	})
```
aaronfern, line 923: If I understood the intention of this, we only need to remove instances of a specific annotation.
If that is the case then a simple delete(map, key) would be sufficient, and `maps.DeleteFunc()` might be an overkill

tools/cmd/scadctl/cmd/genscenario/gardener.go
aaronfern, line 502: Every time we encounter here, the return is `return snap, err`.
Why would we want to return an incomplete snapshot instead of doing a `return nil, err`?

api/planner/types.go
takoverflow: Currently information about the pod owner is inferred from `PodInfo` via the `OwnerReference`. However for any pod that is replicated i.e. has a backing controller, CA attempts to also fetch the corresponding owner for the pod to check if it can be evicted in order to perform scale-in activities.

Currently as part of our ClusterSnapshot, the owner details aren't captured. One could argue that it can be constructed from the OwnerRef, but that doesn't provide details like the `Selector` which is enforced by the API server to be present when constructing objects such as StatefulSet or ReplicaSet.

As for having this information as part of the `PodInfo`, consider that a singular owner can have hundreds of replica pods, each with its own PodInfo, duplicating the owner details such as `Selector` would lead to needless repetition.

<hr>

CA logs when it attempts to evaluate scaledown for a node having pods but where it cannot find the corresponding owner

```
I0616 03:33:24.390925       1 cluster.go:147] Simulating node node-a-2 removal
I0616 03:33:24.390930       1 cluster.go:151] Node node-a-2 cannot be removed: default/half-berry-4 is not replicated
```

```
I0616 04:37:55.606653       1 cluster.go:151] Node node-a-5 cannot be removed: replication controller for default/half-berry-9 is not available, err: replicaset.apps "owner-replica-set" not found
```

Ref: https://github.com/kubernetes/autoscaler/blob/fcdb984e30ac4db88f1df3b30de75ee8e4d8dd76/cluster-autoscaler/simulator/drainability/rules/replicated/rule.go#L46-L48

Ref: https://github.com/kubernetes/autoscaler/blob/fcdb984e30ac4db88f1df3b30de75ee8e4d8dd76/cluster-autoscaler/simulator/drainability/rules/replicacount/rule.go#L49

api/planner/types.go
takoverflow: APIserver enforces this requirement, creating an owner object with empty selector gives errors like

```
Error: error deploying scheduled pods: failed to create replicaset "default/owner-replica-set": ReplicaSet.apps "owner-replica-set" is invalid: [spec.selector: Required value, spec.template.metadata.labels: Invalid value: null: `selector` does not match template `labels`, spec.template.spec.containers: Required value]
```

common/volutil/volutil.go
takoverflow: Deploying objects with non-empty `ResourceVersion` results in APIserver errors.

tools/cmd/scadctl/cmd/genscenario/gardener.go
takoverflow: I added the motivation behind adding the `PodOwner` struct in its docstring

> // This data is currently used by the benchmarking harness to reconstruct the
// pod owners, allowing cluster-autoscaler to perform scale-in correctly.

I should've maybe made it more explicit why it was introduced in the PR description.

takoverflow: I'm trying to analyse if it's possible to recreate the required information from the Pod OwnerRef and create a dummy `Owner` object for CA. It that works out, we can bypass the need for having `OwnerInfo`. cc: @elankath

api/planner/types.go
takoverflow: CA ignores DaemonSet pods entirely, it relies on the actual DaemonSet object being there to compute DS pods' utilization.

Internally in a discussion, it was discussed that the advisor could also do a similar optimization however that was parked for later.

tools/cmd/scadctl/cmd/genscenario/gardener.go
takoverflow: If an error is returned, regardless of whether `snap` was `nil` or an incomplete snapshot, it's never consumed.

api/planner/types.go
takoverflow: Not strictly necessary, both `CurrentReplicas` and `TargetReplicas` were added to allow CA to have the necessary data it required for injecting recently evicted pods.

Ref: https://github.com/kubernetes/autoscaler/blob/fcdb984e30ac4db88f1df3b30de75ee8e4d8dd76/cluster-autoscaler/core/scaledown/planner/planner.go#L262

However this is not necessary from a functional PoV and can be removed. 👍

api/planner/types.go
takoverflow: After offline discussion, it was decided that the `PodOwnerInfo` object would be removed in favour of creating a dummy Owner object on the fly using the pod `OwnerRef` to avoid CA scale-down from being blocked.

tools/cmd/scadctl/cmd/genscenario/gardener.go
takoverflow: Removed with https://github.com/gardener/scaling-advisor/pull/164/commits/6571e8c66abcf33cf551f25f900cfd06b00344c4

api/planner/types.go
takoverflow: Removed with https://github.com/gardener/scaling-advisor/pull/164/commits/6571e8c66abcf33cf551f25f900cfd06b00344c4

api/planner/types.go
takoverflow: Removed with https://github.com/gardener/scaling-advisor/pull/164/commits/6571e8c66abcf33cf551f25f900cfd06b00344c4

PR #163, opened 2026-05-14
Scale in planner

r4mek:
**What this PR does / why we need it**:
This PR introduces scale-in planner.

**Which issue(s) this PR fixes**:
Fixes #31 
Fixes #99 
Fixes #100 
Fixes #149 
Fixes #150 

**Special notes for your reviewer**:

**Release note**:

```feature user

```

PR #168, opened 2026-05-29
API Changes for Operator

elankath:
This PR introduces a set of API changes to `ScalingConstraint`, `ScalingAdvice`, and `ScalingFeedback`. The following changes have been made

1. `NodeTemplates` are moved from individual `NodePool` entries to the top level of the `ScalingConstraint` spec. This makes them reusable across node pools. 
    To associate node templates to node pools,  `NodePool` now contain a `Requirements []NodePoolRequirement` field that uses a selector-based approach to match them.
2. `NodeTemplate`s do not contain a priority field anymore. Instead priority is not defined as part of the `Requirement` in `nodePool.Requirement`
3. `ScalingFeedback` is no longer a standalone CRD resource. It is now part of the `ScalingAdvice.Status`.
4. The following fields have been renamed in `ScalingAdviceSpec`
  i. `ScaleOutPlan` -> `ScaleOut`
  ii. `ScaleInPlan` -> `ScaleIn`
5. A new utility package `common/scalingconstraintutil` has been introduced to provide helper methods for the matching logic between node pools and node templates
6. The following fields have been removed from `ScalingConstraintSpec`
  i. `DefaultBackoffPolicy`
  ii. `ScaleInPolicy`
  iii. `ConsumerID`
7. The following fields have been removed from `NodePool`
  i. `NodeTemplates`
  ii. `ScaleInPolicy`
  iii. `BackoffPolicy`

Fixes #

Validation is not added as part of this PR. I intend on adding all validation as part of another separate PR

**Release note**:

```feature user

```


planner/simulator/scaleout/scaleout.go
```diff
@@ -175,20 +177,63 @@ func SendPlanResult(ctx context.Context, resultCh chan<- plannerapi.ScaleOutPlan
...
+			if !found || req.Priority > best.Priority {
+				best = req
+				found = true
+			}
+		}
+	}
+	return best, found
+}
+
+func nodeTemplateMatchesRequirement(nt sacorev1alpha1.NodeTemplate, req sacorev1alpha1.NodePoolRequirement) bool {
```
takoverflow: This already exists in `scalingconstraintutil`, can be removed from here.

api/core/v1alpha1/scalingconstraints.go
```diff
@@ -158,16 +177,3 @@ type InstancePricing struct {
...
-// BackoffPolicy defines the backoff policy to be used when backing off from suggesting an instance type + zone in subsequence scaling advice upon failed scaling operation.
-type BackoffPolicy struct {
-	// InitialBackoffDuration defines the lower limit of the backoff duration.
-	InitialBackoffDuration metav1.Duration `json:"initialBackoff"`
-	// MaxBackoffDuration defines the upper limit of the backoff duration.
-	MaxBackoffDuration metav1.Duration `json:"maxBackoff"`
-}
-
-// ScaleInPolicy defines the scale in policy to be used when scaling in a node pool.
-type ScaleInPolicy struct {
```
takoverflow, line 171: Why was `ScaleInPolicy` removed? Will this be added back when #163 is merged?

api/core/v1alpha1/scalingconstraints.go
```diff
@@ -40,18 +40,10 @@ type ScalingConstraintList struct {
...
-	// DefaultBackoffPolicy defines a default backoff policy for all NodePools of a cluster. Backoff policy can be overridden at the NodePool level.
-	// +optional
-	DefaultBackoffPolicy *BackoffPolicy `json:"defaultBackoffPolicy,omitempty"`
-	// ScaleInPolicy defines the default scale in policy to be used when scaling in a node pool.
-	// +optional
-	ScaleInPolicy *ScaleInPolicy `json:"scaleInPolicy,omitempty"`
-	// ConsumerID is the Name of the consumer who creates the scaling constraint and is the target for cluster scaling advice.
-	// It allows a consumer to accept or reject the advice by checking the ConsumerID for which the scaling advice has been created.
-	// +optional
-	ConsumerID string `json:"consumerID,omitempty"`
```
takoverflow, line 52: Is the reason behind removal of `ConsumerID` that the `ObjectMeta` for the `ScalingConstraint` object is sufficient to associate a `ScalingAdvice` with a Constraint?

api/core/v1alpha1/scalingconstraints.go
```diff
@@ -40,18 +40,10 @@ type ScalingConstraintList struct {
...
 	// NodePools is the list of node pools to choose from when creating scaling advice.
 	NodePools []NodePool `json:"nodePools,omitempty"`
+	// NodeTemplates is the slice of all NodeTemplates used within the scaling constraint spec.
```
takoverflow:
```suggestion
	// NodeTemplates is the slice of all NodeTemplates that can be used for selecting instances associated with each NodePool.
```

api/core/v1alpha1/scalingfeedback.go
```diff
@@ -55,20 +33,37 @@ const (
...
-	// AvailabilityZone is the availability zone of the node pool.
-	AvailabilityZone string `json:"availabilityZone"`
-	// InstanceType is the instance type of the node pool.
-	InstanceType string `json:"instanceType"`
+// ScaleOutFeedback is the feedback from the life cycle manager when applying an [ScaleOutPlan] of a [ScalingAdviceSpec]
+type ScaleOutFeedback struct {
+	Items []ScaleOutItemFeedback `json:"items,omitempty"`
+}
+
+// ScaleOutItemFeedback is the feedback from the life cycle manager when applying an individual [ScaleOutItem]
```
takoverflow, line 37: Should the doc-string also elaborate on various stages of processing of the `ScaleOutItem` by the lifecycle manager and how it relates to `ScaleOutItemFeedback`? That is:
1. An advice item is considered as `acknowledged` when the lifecycle manager populates `CreationDeadline`
2. It can be considered as `rejected` if `ErrorType` and `BackoffUntil` are populated.
3. Partially applied if `FailCount` is non-zero.

The only question is, how is `Successfully applied` indicated? How to distinguish between `Acknowledged` and `Success` stages?

api/core/v1alpha1/scalingfeedback.go
```diff
@@ -55,20 +33,37 @@ const (
...
+	// +required
+	CreationDeadline metav1.Time `json:"creationDeadline"`
+	// BackoffUtil if populated, represents the time until the scaling-advisor will not consider the corresponding
+	// [ScaleOutItem]'s [NodePlacement] when running simulations and generating subsequent [ScaleOutPlan]'s
+	// +optional
+	BackoffUntil *metav1.Time `json:"backoffUntil,omitempty"`
 	// ErrorType is the type of error that occurred during scale-out.
-	ErrorType ScalingErrorType `json:"errorType"`
+	// +optional
+	ErrorType ScalingErrorType `json:"errorType,omitempty"`
```
takoverflow, line 51: If it only consists of errors during scale-out, wouldn't it be more appropriate to call it `ScaleOutErrorType` rather than the generic `ScalingErrorType`?

api/core/v1alpha1/validation.go
```diff
@@ -18,16 +18,19 @@ func ValidateNodePool(np *NodePool, fldPath *field.Path) (allErrs field.ErrorLis
 	if len(np.AvailabilityZones) == 0 {
 		allErrs = append(allErrs, field.Required(fldPath.Child("availabilityZones"), "availabilityZone must not be empty"))
 	}
-	if len(np.NodeTemplates) == 0 {
```
takoverflow, line 21: Shouldn't the validation now check that the `Requirements` specified for the pool should be non-empty?

Or is that a valid requirement, indicating no restriction over the chosen instance types i.e. all nodeTemplates are part of such a nodePool?

api/core/v1alpha1/validation.go
```diff
@@ -18,16 +18,19 @@ func ValidateNodePool(np *NodePool, fldPath *field.Path) (allErrs field.ErrorLis
...
 	if np.Priority < 0 {
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("priority"), np.Priority, "priority must be non-negative"))
 	}
 	// TODO add checks for Quota
 	return allErrs
 }
 
+// ValidateNodeTemplate validates a NodeTemplate object.
+func ValidateNodeTemplate(_ *NodeTemplate, _ *field.Path) (allErrs field.ErrorList) {
+	//TODO: add validation for NodeTemplate
```
takoverflow: Is this planned for a later PR?

common/scalingconstraintutil/scalingconstraintutil.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"slices"
+
+	sacorev1alpha1 "github.com/gardener/scaling-advisor/api/core/v1alpha1"
+)
+
+// NodePoolToNodeTemplates returns a map of NodePool.Name to the matching NodeTemplates
+// from sc.Spec.NodeTemplates, based on each pool's Requirements.
+func NodePoolToNodeTemplates(sc *sacorev1alpha1.ScalingConstraint) map[string][]sacorev1alpha1.NodeTemplate {
+	result := make(map[string][]sacorev1alpha1.NodeTemplate, len(sc.Spec.NodePools))
```
takoverflow: Consider renaming `result` to `poolTemplateMap` instead?

common/scalingconstraintutil/scalingconstraintutil.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func nodeTemplateMatchesRequirement(nt sacorev1alpha1.NodeTemplate, req sacorev1alpha1.NodePoolRequirement) bool {
+	var val string
+	switch req.Key {
+	case "node.kubernetes.io/instance-type":
+		val = nt.InstanceType
+	case "kubernetes.io/arch":
+		val = nt.Architecture
+	default:
```
takoverflow, line 45: Should a warning or something be emitted when unsupported keys are added to the requirement?

If we're only supporting specifying these two fields (instance-type and arch) as requirement keys, maybe the validation can be strengthened to restrict only specifying those for now, can be relaxed in the future when support for more req keys is added. WDYT?

Are we not supporting capacity based requirements?

common/scalingconstraintutil/scalingconstraintutil.go
```diff
@@ -0,0 +1,81 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+func nodeTemplateMatchesRequirement(nt sacorev1alpha1.NodeTemplate, req sacorev1alpha1.NodePoolRequirement) bool {
+	var val string
+	switch req.Key {
+	case "node.kubernetes.io/instance-type":
```
takoverflow:
```suggestion
	case corev1.LabelInstanceTypeStable:
```

common/scalingconstraintutil/scalingconstraintutil.go
```diff
@@ -0,0 +1,81 @@
...
+		val = nt.InstanceType
+	case "kubernetes.io/arch":
```
takoverflow:
```suggestion
	case corev1.LabelArchStable:
```

api/core/v1alpha1/scalingconstraints.go
```diff
@@ -74,49 +66,78 @@ type ScalingConstraintStatus struct {
...
+	// +required
+	Key string `json:"key"`
+	// Operator represents a key's relationship to a set of values.
+	// Valid operators are In, NotIn, Exists, DoesNotExist. Gt, and Lt.
+	// +required
+	Operator NodePoolRequirementOperator `json:"operator"`
+	// Values is an array of string values. If the operator is "In" or "NotIn",
+	// the values array must be non-empty. If the operator is "Exists" or "DoesNotExist:,
+	// the values array must be empty. If the operator is "Gt" or "Lt", the values
+	// array must have a single element, which will be interpreted as an integer.
```
takoverflow, line 111: Is this the case? If I want to specify capacity based requirements i.e. `cpu > 500m` or `memory < 6Gi`, do I have to specify the values as integers?

common/scalingconstraintutil/scalingconstraintutil.go
```diff
@@ -0,0 +1,81 @@
...
+	switch req.Operator {
+	case sacorev1alpha1.NodePoolRequirementOpIn:
+		return slices.Contains(req.Values, val)
+	case sacorev1alpha1.NodePoolRequirementOpNotIn:
+		return !slices.Contains(req.Values, val)
+	case sacorev1alpha1.NodePoolRequirementOpExists:
+		return val != ""
+	case sacorev1alpha1.NodePoolRequirementOpDoesNotExist:
+		return val == ""
+	default:
```
takoverflow, line 58: What about `Lt` or `Gt` requirements?

common/scalingconstraintutil/scalingconstraintutil.go
```diff
@@ -0,0 +1,81 @@
...
+		return false
+	}
+}
+
+// GetNodePlacements computes and returns all the possible `NodePlacement`s for this NodePool.
+func GetNodePlacements(sc sacorev1alpha1.ScalingConstraint) map[string][]sacorev1alpha1.NodePlacement {
+	result := make(map[string][]sacorev1alpha1.NodePlacement, len(sc.Spec.NodePools))
```
takoverflow:
```suggestion
	poolPlacementMap := make(map[string][]sacorev1alpha1.NodePlacement, len(sc.Spec.NodePools))
```

planner/simulator/scaleout/scaleout.go
```diff
@@ -175,20 +177,63 @@ func SendPlanResult(ctx context.Context, resultCh chan<- plannerapi.ScaleOutPlan
...
 			for _, az := range np.AvailabilityZones {
-				allNodeTemplates = append(allNodeTemplates, createNodeTemplate(np, nt, az))
+				allNodeTemplates = append(allNodeTemplates, createNodeTemplate(np, nt, az, matchingReq.Priority))
 			}
 		}
 	}
 	return allNodeTemplates
 }
 
+func findMatchingRequirement(nt sacorev1alpha1.NodeTemplate, requirements []sacorev1alpha1.NodePoolRequirement) (sacorev1alpha1.NodePoolRequirement, bool) {
```
takoverflow: Shouldn't this just return the highest priority and if nothing's found just return 0? We're anyways ignoring the found bool and only consider the priority from the returned requirement.

planner/simulator/scaleout/scaleout.go
```diff
@@ -175,20 +177,63 @@ func SendPlanResult(ctx context.Context, resultCh chan<- plannerapi.ScaleOutPlan
...
+	var best sacorev1alpha1.NodePoolRequirement
+	found := false
+	for _, req := range requirements {
+		if nodeTemplateMatchesRequirement(nt, req) {
+			if !found || req.Priority > best.Priority {
```
takoverflow: The `found` bool and its logic can be removed.

common/scalingconstraintutil/scalingconstraintutil.go
```diff
@@ -0,0 +1,81 @@
...
+		return val != ""
+	case sacorev1alpha1.NodePoolRequirementOpDoesNotExist:
+		return val == ""
+	default:
+		return false
+	}
+}
+
+// GetNodePlacements computes and returns all the possible `NodePlacement`s for this NodePool.
+func GetNodePlacements(sc sacorev1alpha1.ScalingConstraint) map[string][]sacorev1alpha1.NodePlacement {
```
takoverflow, line 64: Shouldn't this be part of planner testutil only? I don't see it being used anywhere else.

tools/cmd/scadctl/cmd/genscenario/gardener.go
```diff
@@ -573,18 +576,20 @@ func constructInstanceRequirementsMap(instances []gardenercorev1beta1.MachineTyp
...
 	// TODO csc.Spec.ConsumerID = "abcd", backoffpolicy, scaleinpolicy
 	return
 }
 
-func (a *access) createNodePools() (nodePools []sacorev1alpha1.NodePool) {
+func (a *access) createNodePoolsAndNodeTemplates() (nodePools []sacorev1alpha1.NodePool, nodeTemplates []sacorev1alpha1.NodeTemplate) {
 	region := a.shoot.Spec.Region
 
 	for _, worker := range a.shoot.Spec.Provider.Workers {
 		var nodePool sacorev1alpha1.NodePool
```
takoverflow, line 591: Where are the pool requirements being added?

takoverflow (commented):
Thanks for the changes, the switch to flexible requirements in the pool would allow for more expressive customizability for the workers!

I've added a couple of comments PTAL

samples/data/scaling-constraints/1p.yaml
```diff
@@ -5,18 +5,24 @@ spec:
...
-          name: m5l
+      region: eu-west-1
+      priority: 0
+      requirements:
+        - key: "node.kubernetes.io/instance-type"
+          operator: In
+          values:
+            - m5.large
           priority: 0
       priority: 0
```
gagan16k:
```suggestion
```
Causing formatting error: `Map keys must be unique`

common/scalingconstraintutil/scalingconstraintutil.go
```diff
@@ -0,0 +1,81 @@
...
+			if nodeTemplateMatchesRequirements(nt, np.Requirements) {
+				result[np.Name] = append(result[np.Name], nt)
+			}
+		}
+	}
+	return result
+}
+
+func nodeTemplateMatchesRequirements(nt sacorev1alpha1.NodeTemplate, requirements []sacorev1alpha1.NodePoolRequirement) bool {
+	for _, req := range requirements {
```
gagan16k, line 29: Since `NodePool.Requirements` is +optional, a pool with no `requirements` will inherit every template in the spec, as it would return `true`.

planner/simulator/scaleout/scaleout.go
```diff
@@ -175,20 +177,63 @@ func SendPlanResult(ctx context.Context, resultCh chan<- plannerapi.ScaleOutPlan
...
+
+func nodeTemplateMatchesRequirement(nt sacorev1alpha1.NodeTemplate, req sacorev1alpha1.NodePoolRequirement) bool {
+	var val string
+	switch req.Key {
+	case "node.kubernetes.io/instance-type":
+		val = nt.InstanceType
+	case "kubernetes.io/arch":
+		val = nt.Architecture
+	default:
+		return false
```
gagan16k: Needs a log message to prevent silently dropping templates

api/core/v1alpha1/scalingfeedback.go
```diff
@@ -55,20 +33,37 @@ const (
...
-// ScaleInErrorInfo is the information about nodes that could not be deleted for scale-in.
-type ScaleInErrorInfo struct {
-	// NodeNames is the list of node names that could not be deleted for scaled in.
-	NodeNames []string `json:"nodeNames"`
+// ScaleInFeedback is  the feedback from the life cycle manager after applying [ScaleInPlan]
+type ScaleInFeedback struct {
+	// AcceptedNodeNames holds the slice of node names that were accepted for scale-in by the lifecycle controller.
+	// Required to be specified, since if empty, the ScaleInFeedback itself should not be populated.
+	// +required
+	AcceptedNodesNames []string `json:"acceptedNodesNames"`
```
gagan16k:
```suggestion
	AcceptedNodeNames []string `json:"acceptedNodeNames"`
```
nit

api/core/v1alpha1/scalingfeedback.go
```diff
@@ -55,20 +33,37 @@ const (
...
+}
+
+// ScaleOutItemFeedback is the feedback from the life cycle manager when applying an individual [ScaleOutItem]
+type ScaleOutItemFeedback struct {
+	// CreationDeadline represents the time after which the scaling-advisor can expect real nodes to be created and available
+	// for the corresponding [ScaleOutItem]'s [NodePlacement]. When the [ScalingFeedback] is constructed by the life-cycle manager,
+	// this field is mandatory to be set inside all [ScaleOutItemFeedback]
+	// +required
+	CreationDeadline metav1.Time `json:"creationDeadline"`
+	// BackoffUtil if populated, represents the time until the scaling-advisor will not consider the corresponding
```
gagan16k:
```suggestion
	// BackoffUntil if populated, represents the time until the scaling-advisor will not consider the corresponding
```
nit

api/core/v1alpha1/scalingconstraints.go
```diff
@@ -74,49 +66,78 @@ type ScalingConstraintStatus struct {
...
+type NodePoolRequirement struct {
+	// Key is the label key that the selector applies to.
+	// +required
+	Key string `json:"key"`
+	// Operator represents a key's relationship to a set of values.
+	// Valid operators are In, NotIn, Exists, DoesNotExist. Gt, and Lt.
+	// +required
+	Operator NodePoolRequirementOperator `json:"operator"`
+	// Values is an array of string values. If the operator is "In" or "NotIn",
+	// the values array must be non-empty. If the operator is "Exists" or "DoesNotExist:,
```
gagan16k:
```suggestion
	// the values array must be non-empty. If the operator is "Exists" or "DoesNotExist,
```
nit

api/core/v1alpha1/scalingconstraints.go
```diff
@@ -40,18 +40,10 @@ type ScalingConstraintList struct {
...
-	// +optional
-	ScaleInPolicy *ScaleInPolicy `json:"scaleInPolicy,omitempty"`
-	// ConsumerID is the Name of the consumer who creates the scaling constraint and is the target for cluster scaling advice.
-	// It allows a consumer to accept or reject the advice by checking the ConsumerID for which the scaling advice has been created.
-	// +optional
-	ConsumerID string `json:"consumerID,omitempty"`
 	// NodePools is the list of node pools to choose from when creating scaling advice.
 	NodePools []NodePool `json:"nodePools,omitempty"`
+	// NodeTemplates is the slice of all NodeTemplates used within the scaling constraint spec.
+	NodeTemplates []NodeTemplate `json:"nodeTemplates"`
```
gagan16k, line 48: Should we have a `+required` tag on this?

planner/simulator/scaleout/scaleout.go
```diff
@@ -175,20 +177,63 @@ func SendPlanResult(ctx context.Context, resultCh chan<- plannerapi.ScaleOutPlan
...
+	case sacorev1alpha1.NodePoolRequirementOpDoesNotExist:
+		return val == ""
+	default:
+		return false
+	}
+}
+
 // GroupScaleOutNodeTemplatesByPriority does just exactly that and returns a map keyed by PriorityKey to slice of
 // [plannerapi.ScaleOutNodeTemplate]
 func GroupScaleOutNodeTemplatesByPriority(templates []plannerapi.ScaleOutNodeTemplate) map[commontypes.PriorityKey][]plannerapi.ScaleOutNodeTemplate {
```
gagan16k, line 212: Not a part of this PR, but this function seems to be double-counting the first element of each new group, once when the group is initialized `group = []ScaleOutNodeTemplate{t}` and again by the unconditional `append(group, t)`

common/scalingconstraintutil/scalingconstraintutil.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"slices"
+
+	sacorev1alpha1 "github.com/gardener/scaling-advisor/api/core/v1alpha1"
+)
+
+// NodePoolToNodeTemplates returns a map of NodePool.Name to the matching NodeTemplates
+// from sc.Spec.NodeTemplates, based on each pool's Requirements.
+func NodePoolToNodeTemplates(sc *sacorev1alpha1.ScalingConstraint) map[string][]sacorev1alpha1.NodeTemplate {
```
gagan16k: Can this pass just the spec instead of the entire constraint, as we only use the spec in the function?

gagan16k (commented):
Thanks for picking up this PR! Have some comments, PTAL

api/core/v1alpha1/scalingconstraints.go
aaronfern: ScaleInPolicy was never really designed. It was just left as a TODO stub.
I also see that #163 doesn't really use `ScaleInPolicy`. Any change that requires it can reintroduce it

api/core/v1alpha1/scalingconstraints.go
aaronfern: Yes. 
`ScalingAdvice`s will be produced for and tied to a `ScalingConstraint` object. The metadata is sufficient for this.

planner/simulator/scaleout/scaleout.go
aaronfern: Thanks for pointing this out. Removed.

api/core/v1alpha1/scalingfeedback.go
aaronfern: I would rather not do that here. I think it's better to do it with the operator as the nuances of this might change.
I'll add a TODO here to have this in with the operator

api/core/v1alpha1/scalingfeedback.go
aaronfern: I would rather have the generic type name here. It does not affect readability and makes things more general

api/core/v1alpha1/validation.go
aaronfern: Yes, I will leave validations to be done at a later stage.
These validating functions are not used anywhere for now. Will do all required validation together in a separate PR

common/scalingconstraintutil/scalingconstraintutil.go
aaronfern: > If we're only supporting specifying these two fields (instance-type and arch) as requirement keys, maybe the validation can be strengthened to restrict only specifying those for now, can be relaxed in the future when support for more req keys is added. WDYT?

Yes, will add validations for this in a separate PR

> Are we not supporting capacity based requirements?

Maybe not right now. We can always extend this for other requirement types in the coming weeks

common/scalingconstraintutil/scalingconstraintutil.go
aaronfern: Currently with Instance type and arch type, `Lt` and `Gt` do not make sense. Support for these can be added when other requirement types are supported.

planner/simulator/scaleout/scaleout.go
aaronfern: Updated this to consider the bool in the caller and continue the loop all the way to find the highest priority

common/scalingconstraintutil/scalingconstraintutil.go
aaronfern: You're right, it can be moved to planner testutils.
I didn't do that because this function is a utility of scaling constraints and can be considered for use by anyone else. 
If you have a strong opinion on not having this here because it is only used by the planner utils then I can move it. WDYT?

planner/simulator/scaleout/scaleout.go
aaronfern: Good catch
Added else block to stop double counting

planner/simulator/scaleout/scaleout.go
```diff
@@ -175,20 +176,41 @@ func SendPlanResult(ctx context.Context, resultCh chan<- plannerapi.ScaleOutPlan
...
 	}
 	return allNodeTemplates
 }
 
+func findHighestPrioMatchingRequirement(nt sacorev1alpha1.NodeTemplate, requirements []sacorev1alpha1.NodePoolRequirement) (sacorev1alpha1.NodePoolRequirement, bool) {
+	var best sacorev1alpha1.NodePoolRequirement
+	found := false
+	for _, req := range requirements {
+		if scalingconstraintutil.NodeTemplateMatchesRequirement(nt, req) {
+			if !false || req.Priority >= best.Priority {
```
takoverflow:
```suggestion
			if req.Priority >= best.Priority {
```
`!false` is always true

<hr>

Alternatively, Why are we returning an entire `NodePoolRequirement` when the called only uses `matchingReq.Priority`?

Couldn't this be simplified to
```go
func findHighestPrioMatchingRequirement(nt sacorev1alpha1.NodeTemplate, requirements []sacorev1alpha1.NodePoolRequirement) int32 {
	highestPrio := -1
	for _, req := range requirements {
		if scalingconstraintutil.NodeTemplateMatchesRequirement(nt, req) {
			highestPrio = max(req.Priority, highestPrio)
		}
	}
	return highestPrio
}
```

And the caller can just continue if the returned priority is `-1`, or use the returned priority for creating a nodeTemplate.

No need for a bool, no need to return a dummy requirement.

aaronfern: > !false is always true

🤦🏻  My apologies. Stupid mistake

> Alternatively, Why are we returning an entire NodePoolRequirement when the called only uses matchingReq.Priority?

Yup, you're right. Updated `findHighestPrioMatchingRequirement`

common/scalingconstraintutil/scalingconstraintutil.go
```diff
@@ -0,0 +1,83 @@
...
+	corev1 "k8s.io/api/core/v1"
+)
+
+// NodePoolToNodeTemplates returns a map of NodePool.Name to the matching NodeTemplates
+// from sc.Spec.NodeTemplates, based on each pool's Requirements.
+func NodePoolToNodeTemplates(sc *sacorev1alpha1.ScalingConstraintSpec) map[string][]sacorev1alpha1.NodeTemplate {
+	poolTemplateMap := make(map[string][]sacorev1alpha1.NodeTemplate, len(sc.NodePools))
+	for _, np := range sc.NodePools {
+		for _, nt := range sc.NodeTemplates {
+			if nodeTemplateMatchesRequirements(nt, np.Requirements) {
```
gagan16k, line 20: This will be O(`pool` x `templates` x `requirements`), could be optimised. Would be worth doing so, as this function could be useful in a lot of places.

planner/testutil/testutil.go
```diff
@@ -331,8 +332,9 @@ func (s *testStorageMetaAccess) GetFallbackCSINodeSpec(instanceType string) (csi
 // getAllNodePlacements computes all the possible NodePlacements for the ScalingConstraintSpec.
 func getAllNodePlacements(c sacorev1alpha1.ScalingConstraintSpec) (placements []sacorev1alpha1.NodePlacement) {
 	placements = make([]sacorev1alpha1.NodePlacement, 0, len(c.NodePools)*len(c.GetAllAvailabilityZones()))
-	for _, p := range c.NodePools {
-		placements = append(placements, p.GetNodePlacements()...)
+	allNodePlacements := scalingconstraintutil.GetNodePlacements(sacorev1alpha1.ScalingConstraint{Spec: c})
+	for _, p := range allNodePlacements {
+		placements = append(placements, p...)
```
gagan16k:
```suggestion
	allNodePlacements := scalingconstraintutil.GetNodePlacements(sacorev1alpha1.ScalingConstraint{Spec: c})
	for _, p := range c.NodePools {
		placements = append(placements, allNodePlacements[p.Name]...)
```

The iteration order for golang maps are [non-deterministic](https://go.dev/blog/maps#iteration-order). This change could reduce flake in tests, especially when a particular order is expected in tests (like `planner_test.go:306`)

api/core/v1alpha1/scalingfeedback.go
```diff
@@ -10,38 +10,16 @@ import (
...
-
-// ScalingFeedbackSpec defines the specification of the ScalingFeedback.
-type ScalingFeedbackSpec struct {
+	// ScaleOut is the scale-out feedback from the lifecycle manager when applying [ScaleOutPlan]
+	// [ScalingAdviceSpec].
+	ScaleOut *ScaleOutFeedback `json:"scaleOut,omitempty"`
+	// ScaleIn is the scale-in feedback from life-cycle manager when applying [ScaleInPlan]
+	ScaleIn *ScaleInFeedback `json:"scaleIn,omitempty"`
 	// ConstraintRef is a reference to the ScalingConstraint that this advice is based on.
 	ConstraintRef apicommon.NamespacedName `json:"constraintRef"`
```
gagan16k: Might be redundant, as the enclosing ScalingAdvice.Spec already has it.

common/scalingconstraintutil/scalingconstraintutil.go
takoverflow: Alright, we can keep it as is.

tools/cmd/scadctl/cmd/genscenario/gardener.go
takoverflow: @aaronfern Can you add a `TODO` here under my name. I can address this as part of #164 (that PR would anyways require modifications once this is merged)

aaronfern: Pushed change to construct nodePool.Requirements in https://github.com/gardener/scaling-advisor/pull/168/changes/eaa5c70f8d9e65708f09c6d865cf72e797aa6a6f
PTAL. If you're still unsure and would still like to evaluate in #164, I can still add the `TODO` comment

takoverflow: This should be fine for now, thanks!

api/core/v1alpha1/scalingfeedback.go
aaronfern: Yup, makes sense. Removed in https://github.com/gardener/scaling-advisor/pull/168/changes/83386773fefd9e62d605e5f6ff4d13e62cc36417

common/scalingconstraintutil/scalingconstraintutil.go
takoverflow: How do you propose this optimization be performed? I don't see any way of skipping some template or requirements from being checked.

PR #171, opened 2026-06-12
Minkapi view isolation fix

r4mek:
This PR fixes the issue where the objects only present in baseview were getting deleted when a delete request was issued to sandboxview. This was not desirable as view isolation between baseview and sandboxview was broken.

Fixes #

Now we have introduced tombstone markers in views to indicate that the object has been deleted.

**Release note**:

```feature user

```


minkapi/view/sandbox.go
```diff
@@ -44,7 +45,12 @@ type sandboxView struct {
 	args         *minkapi.ViewArgs
 	mu           *sync.RWMutex
 	stores       map[schema.GroupVersionKind]*store.InMemResourceStore
-	changeCount  atomic.Int64
+	// tombstones records, per GVK, the namespaced names of objects that have been deleted
+	// from this sandbox's perspective. Tombstoned names cause Get/List/Watch on this view to
+	// behave as if the object is gone, while the delegate's copy (if any) is left untouched.
+	// Cleared on Reset and on a successful CreateObject of the same name.
+	tombstones  map[schema.GroupVersionKind]map[string]struct{}
```
elankath: The entire point of spending time doing that `cache.DeletedFinalStateUnknown` research yesterday was to precisely avoid maintaining separate `tombstones` map and instead have tombstone managed as a feature of `ResourceStore`.

Please do not maintain separate map here. Introduce necessary methods in `ResourceStore`: `MarkDelete..` and `IsMarkedForDelete` and have tombstone handling be done natively within `ResourceStore`

r4mek: Will do the changes.

r4mek: I was wondering, if instead of having `MarkForDelete()`, I can use `ObjectOption` to have something like `LogicalDelete`  and I don't really need `IsMarkedForDelete()`, I can just check for `ErrObjectDeleted` to understand that it has been marked for delete. This would clean up the interface a bit. WDYT?

@elankath

api/minkapi/types.go
```diff
@@ -92,6 +92,15 @@ type ResourceStore interface {
 	GetWatcher(ctx context.Context, namespace string, options metav1.ListOptions) (watch.Interface, error)
 	// GetVersionCounter returns the atomic counter for generating monotonically increasing resource versions
 	GetVersionCounter() *atomic.Int64
+	// MarkForDelete records a tombstone for the given key: any prior live object at that key is
+	// removed from the store, a Deleted watch event is broadcast (carrying the typed object as
+	// payload, not the wrapper), and a [k8s.io/client-go/tools/cache.DeletedFinalStateUnknown]
```
takoverflow: > carrying the typed object as payload, not the wrapper

What does this mean?

api/minkapi/types.go
```diff
@@ -98,6 +99,8 @@ type ResourceStore interface {
 type ObjectOptions struct {
 	// NoBroadcast means there will be no watch event broadcasted when this object is created or updated.
 	NoBroadcast bool
+	// LogicalDelete means that the object would be replaced by tombstone; else we delete the object from view.
+	LogicalDelete bool
```
takoverflow: What does the prefix `Logical` indicate here?

minkapi/view/testutil/testutil.go
```diff
@@ -0,0 +1,100 @@
...
+	PatchPodA    = "patch-pod-a.json"
+	PodA         = "pod-a.json"
+	PodDefaultNS = "pod-defaultns.json"
+	PodTestNS    = "pod-testns.json"
+	UidTSPodA    = "uid-ts-pod-a.json"
+	UpdatePodA   = "update-pod-a.json"
+)
+
+var (
+	defaultBaseViewArgs = &mkapi.ViewArgs{
```
takoverflow: Is there any need to declare the `defaultBaseViewArgs` and `defaultSandboxViewArgs` here? Why can't they directly be returned inside the `GetDefaultBaseViewArgs()` and `GetDefaultSandboxViewArgs()` itself? (they're not being used anywhere else)

Alternatively, they can be exported variables and you can use them directly rather than having these `Get`ter functions

minkapi/view/testutil/testutil.go
```diff
@@ -0,0 +1,100 @@
...
+		KubeConfigPath: "/tmp/minkapi-test.yaml",
+		Scheme:         typeinfo.SupportedScheme,
+		WatchConfig: mkapi.WatchConfig{
+			QueueSize: mkapi.DefaultWatchQueueSize,
+			Timeout:   mkapi.DefaultWatchTimeout,
+		},
+	}
+	defaultSandboxViewArgs = &mkapi.ViewArgs{
+		Name:           "sandbox",
+		KubeConfigPath: "sandbox",
```
takoverflow: Is this path being used while checking the tests manually? If so, consider providing a proper one.

minkapi/view/testutil/testutil.go
```diff
@@ -0,0 +1,100 @@
...
+	mkapi "github.com/gardener/scaling-advisor/api/minkapi"
+	"github.com/gardener/scaling-advisor/api/minkapi/typeinfo"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+)
+
+const testdataPath = "testutil/testdata"
+
+// Testdata filenames to load objects.
+const (
```
takoverflow, line 19: Now that these testfile names are declared separately from the actual tests that these files correspond to, might be worth having a oneliner for some of these exported constant names that describe what's the modification done to them.

For example, `NameNodeA` doesn't give me any idea how it's different from `NodeA`, unless I go to the test which is using it. Might be worth describing it here as well.

minkapi/view/sandbox.go
```diff
@@ -132,13 +133,20 @@ func (v *sandboxView) GetEventSink() minkapi.EventSink {
 }
 
 func (v *sandboxView) CreateObject(ctx context.Context, gvk schema.GroupVersionKind, obj metav1.Object, opts minkapi.ObjectOptions) (metav1.Object, error) {
+	// A successful Create resurrects a tombstoned name; the underlying store's Add replaces
+	// any DeletedFinalStateUnknown wrapper at the same key with the new object.
```
takoverflow:
```suggestion
	// any DeletedFinalStateUnknown wrapper for the same key with the new object.
```

minkapi/view/store/store.go
```diff
@@ -161,11 +164,19 @@ func (s *InMemResourceStore) DeleteByKey(ctx context.Context, key string) error
 }
 
 // Delete deletes the object identified by its fully qualified cache name from the store. It delegates to DeleteByKey
-func (s *InMemResourceStore) Delete(ctx context.Context, objName cache.ObjectName) error {
+func (s *InMemResourceStore) Delete(ctx context.Context, objName cache.ObjectName, opts minkapi.ObjectOptions) error {
```
takoverflow, line 178: Can you update the doc-string to inform that if the `LogicalDelete` option is passed, the object is not removed from the store, rather its tombstone is just recorded.

minkapi/view/store/store.go
```diff
@@ -336,6 +357,77 @@ func (s *InMemResourceStore) DeleteObjects(ctx context.Context, c minkapi.MatchC
 	return
 }
 
+// markForDelete records a tombstone for the given key. If a live object existed at the key it
+// is removed from the cache, a Deleted watch event is broadcast (carrying the typed object as
+// its payload, NOT the wrapper, to match real apiserver watch semantics), and a
```
takoverflow: > (carrying the typed object as its payload, NOT the wrapper, to match real apiserver watch semantics)

What does this mean?

minkapi/view/store/store.go
```diff
@@ -336,6 +357,77 @@ func (s *InMemResourceStore) DeleteObjects(ctx context.Context, c minkapi.MatchC
...
+// [cache.DeletedFinalStateUnknown] marker is parked at the same key in the underlying
+// cache.Store. Subsequent Get/List/Watch operations treat the key as absent (Get returns
+// [minkapi.ErrObjectDeleted]; List/Watch initial-list skip the entry). Re-Add of the same key
```
takoverflow: What is meant by `initial-list` here?

minkapi/view/store/store.go
```diff
@@ -336,6 +357,77 @@ func (s *InMemResourceStore) DeleteObjects(ctx context.Context, c minkapi.MatchC
...
+		return err
+	}
+	if !exists || isTombstone(obj) {
+		return apierrors.NewNotFound(schema.GroupResource{Group: s.args.ObjectGVK.Group, Resource: s.args.Name}, key)
+	}
+	mo, err := objutil.AsMeta(obj)
+	if err != nil {
+		return err
+	}
+	mo.SetDeletionTimestamp(&metav1.Time{Time: time.Time{}})
```
takoverflow: Shouldn't this be `time.Now()`?

minkapi/view/store/store.go
```diff
@@ -336,6 +357,77 @@ func (s *InMemResourceStore) DeleteObjects(ctx context.Context, c minkapi.MatchC
...
+	if broadcastObj != nil {
+		broadcastObj = broadcastObj.DeepCopyObject()
+	}
+	if broadcastObj != nil {
+		go func() {
+			if berr := s.broadcaster.Action(watch.Deleted, broadcastObj); berr != nil {
+				log.Error(berr, "failed to broadcast object delete", "key", key)
+			}
+		}()
+	}
```
takoverflow: The two `if` statements can be combined in a single one.

minkapi/view/store/store.go
```diff
@@ -336,6 +357,77 @@ func (s *InMemResourceStore) DeleteObjects(ctx context.Context, c minkapi.MatchC
...
+func isTombstone(obj any) bool {
+	switch obj.(type) {
+	case cache.DeletedFinalStateUnknown, *cache.DeletedFinalStateUnknown:
+		return true
+	}
+	return false
+}
+
+// filterTombstones returns a new slice containing only the entries from items that are not
+// tombstones. Used by List/ListMetaObjects/buildPendingWatchEvents/DeleteObjects to hide
```
takoverflow: No need to mention each function that might be using it, this can easily get outdated.
```suggestion
// tombstones. Used to hide tombstoned entries from callers.
```

minkapi/view/store/store.go
```diff
@@ -336,6 +357,77 @@ func (s *InMemResourceStore) DeleteObjects(ctx context.Context, c minkapi.MatchC
...
+// tombstoned entries from callers.
+func filterTombstones(items []any) []any {
```
takoverflow: Consider renaming to `filterOutTombstones`

minkapi/view/store/store.go
```diff
@@ -355,7 +447,7 @@ func (s *InMemResourceStore) buildPendingWatchEvents(startVersion int64, namespa
 	defer s.mu.Unlock()
 
 	var skip bool
-	allItems := s.cache.List()
+	allItems := filterTombstones(s.cache.List())
```
takoverflow: Rather than invoking `filterTombstones` here, consider modifying `shouldSkipObject()` to return `true` for tombstoned objects.

This would allow for the same to be checked even in `s.Watch()` which is not happening right now.

minkapi/view/store/store.go
```diff
@@ -161,11 +164,19 @@ func (s *InMemResourceStore) DeleteByKey(ctx context.Context, key string) error
...
 	return s.DeleteByKey(ctx, objName.String())
 }
 
-// GetByKey gets the object identified by the given key from the store and returns the same as a runtime.Object.
+// GetByKey gets the object identified by the given key from the store and returns the same as
+// a runtime.Object. Three outcomes:
+//   - The key holds a live object → returns the typed runtime.Object, nil error.
+//   - The key was previously deleted → returns nil and a wrapped [minkapi.ErrObjectDeleted]
+//     so callers can distinguish "tombstoned" from "never existed."
+//   - The key was never created → returns nil and a standard apierrors.NotFound.
```
takoverflow:
```suggestion
//   - The key holds a live object -> returns the typed runtime.Object
//   - The key was previously deleted -> returns [minkapi.ErrObjectDeleted]
//     so callers can distinguish tombstoned objects from non-existent ones.
//   - The object for the key was never created -> returns apierrors.NotFound.
```

minkapi/view/sandbox.go
```diff
@@ -132,13 +133,20 @@ func (v *sandboxView) GetEventSink() minkapi.EventSink {
...
 	return storeObject(ctx, v, gvk, obj, opts, &v.changeCount)
 }
 
 func (v *sandboxView) GetObject(ctx context.Context, gvk schema.GroupVersionKind, objName cache.ObjectName) (obj runtime.Object, err error) {
 	obj, err = v.getSandboxObject(ctx, gvk, objName)
+	if err != nil && errors.Is(err, minkapi.ErrObjectDeleted) {
+		// Tombstoned: surface as the standard Kubernetes "not found" error and do not fall
+		// through to the delegate (the sandbox has explicitly deleted this name).
```
takoverflow:
```suggestion
		// through to the delegate (the sandbox has explicitly deleted this object).
```

minkapi/view/sandbox.go
```diff
@@ -148,18 +156,18 @@ func (v *sandboxView) GetObject(ctx context.Context, gvk schema.GroupVersionKind
 func (v *sandboxView) UpdateObject(ctx context.Context, gvk schema.GroupVersionKind, obj metav1.Object, opts minkapi.ObjectOptions) error {
 	objName := objutil.CacheName(obj)
 	sandboxObj, err := v.getSandboxObject(ctx, gvk, objName)
+	if err != nil && errors.Is(err, minkapi.ErrObjectDeleted) {
```
takoverflow: I can see some calls being made to `getSandboxObject` where the returned error is not being checked for equivalence to `ErrObjectDeleted`. For example, in `UpdatePodNodeBinding`. Is that intentional? If not, can all the call sites be updated.

minkapi/view/sandbox.go
```diff
@@ -148,18 +156,18 @@ func (v *sandboxView) GetObject(ctx context.Context, gvk schema.GroupVersionKind
...
 	if sandboxObj != nil { //sandbox object is being updated.
 		return updateObject(ctx, v, gvk, obj, opts, &v.changeCount)
 	}
-	// The object lives only in the delegate. Materialize a deep-copy of the current delegate
-	// state into the sandbox without broadcasting (consumers already saw the original via the
-	// delegate's initial-list, so this materialization is plumbing). Then run the normal
-	// update path against the sandbox copy so the broadcast carries the user-visible change.
-	// WatchObjects' filter suppresses the delegate watcher's events for this object name from
-	// here on, so the sandbox broadcast is the single observable event.
+	// Materialize a deep-copy of the delegate's object into the sandbox store (no broadcast,
```
takoverflow: Is this an intentional comment change? Asking because the details from the last sentence are not present in the change anymore.

minkapi/view/sandbox.go
```diff
@@ -230,6 +238,18 @@ func (v *sandboxView) ListMetaObjects(ctx context.Context, gvk schema.GroupVersi
 	if err != nil {
 		return
 	}
+	// Filter out delegate items whose name has been tombstoned in this sandbox's store.
+	s, _ := v.GetResourceStore(gvk)
+	if len(delegateItems) > 0 {
+		filtered := delegateItems[:0]
+		for _, it := range delegateItems {
```
takoverflow:
```suggestion
		for _, item := range delegateItems {
```

minkapi/view/sandbox.go
```diff
@@ -230,6 +238,18 @@ func (v *sandboxView) ListMetaObjects(ctx context.Context, gvk schema.GroupVersi
...
+			if _, gerr := s.GetByKey(ctx, objutil.CacheName(it).String()); errors.Is(gerr, minkapi.ErrObjectDeleted) {
+				continue
+			}
+			filtered = append(filtered, it)
+		}
+		delegateItems = filtered
+	}
 	if myMax >= delegateMax {
```
takoverflow: Not your changes, but this if-else can be replaced by
```
	maxVersion = max(myMax, delegateMax)
```

minkapi/view/sandbox.go
```diff
@@ -303,34 +328,70 @@ func (v *sandboxView) GetWatcher(ctx context.Context, gvk schema.GroupVersionKin
...
+		if err = s.Delete(ctx, objName, minkapi.ObjectOptions{LogicalDelete: true}); err != nil {
+			return err
+		}
+		v.changeCount.Add(1)
+		return nil
+	} else if errors.Is(gerr, minkapi.ErrObjectDeleted) {
+		// Already tombstoned: NotFound, mirroring K8s API server behavior for repeated delete.
+		return apierrors.NewNotFound(schema.GroupResource{Group: gvk.Group, Resource: gvk.Kind}, objName.String())
+	}
+	// Sandbox doesn't have it; consult delegate.
```
takoverflow, line 334: Before this, shouldn't we check this
```suggestion
	} else if !apierrors.IsNotFound(err) {
         return err
    }
	// Sandbox doesn't have it; consult delegate.
```

minkapi/view/sandbox_test.go
```diff
@@ -459,3 +415,342 @@ func checkNodeInViewIsSame(t *testing.T, v mkapi.View, n *corev1.Node) {
...
+		t.Fatalf("ListNodes failed: %v", err)
+	}
+	for _, n := range nodes {
+		if n.Name == nA.Name {
+			t.Errorf("expected sandbox ListNodes to exclude tombstoned node %q, got %v", nA.Name, nodeNames(nodes))
+		}
+	}
+	hasB := false
+	for _, n := range nodes {
+		if n.Name == nB.Name {
```
takoverflow, line 517: This can be checked as part of the previous loop itself, no need to iterate over the list twice.

minkapi/view/testutil/testutil.go
r4mek: I will export them for others to access.

minkapi/view/testutil/testutil.go
r4mek: Will add those.

minkapi/view/store/store.go
r4mek: Will change it.

minkapi/view/store/store.go
r4mek: Will clean up the comment.

minkapi/view/store/store.go
r4mek: Will do the change.

minkapi/view/store/store.go
r4mek: Will do the change.

minkapi/view/store/store.go
r4mek: Will clean up the go doc

minkapi/view/store/store.go
r4mek: I think the check is redundant, so will remove it.

minkapi/view/sandbox.go
r4mek: That was a miss. Will update it.

minkapi/view/sandbox.go
r4mek: Yes, it was intentional as the details seemed irrelevant.

minkapi/view/sandbox.go
r4mek: Will make the change.

minkapi/view/sandbox.go
r4mek: Will do the change.

minkapi/view/sandbox_test.go
r4mek: Will make the change.

minkapi/view/store/store.go
r4mek: I am doing the filtering in `buildPendingWatchEvents` so `s.Watch()` would only watch on the non-tombstoned objects.

api/minkapi/types.go
r4mek: Renaming the variable to `MarkAsDeleted`.

minkapi/view/testutil/testutil.go
r4mek: I will change the `KubeConfigPath` to `/tmp/minkapi-sandbox-test.yaml`. Similarly, I will change `KubeConfigPath` for baseview to `/tmp/minkapi-base-test.yaml`.

minkapi/view/store/store.go
```diff
@@ -182,6 +199,11 @@ func (s *InMemResourceStore) GetByKey(ctx context.Context, key string) (o runtim
 		err = apierrors.NewNotFound(schema.GroupResource{Group: s.args.ObjectGVK.Group, Resource: s.args.Name}, key)
 		return
 	}
+	if isTombstone(obj) {
+		log.V(6).Info("found tombstone for key", "key", key)
+		err = fmt.Errorf("%w: %s/%s", minkapi.ErrObjectDeleted, s.args.ObjectGVK.Kind, key)
```
elankath: We should return `NotFound` wrapping `ErrObjectDeleted`. (Probably easier the other way around) so that `errors.Is` passes for both error codes.

minkapi/view/store/store.go
```diff
@@ -160,12 +167,22 @@ func (s *InMemResourceStore) DeleteByKey(ctx context.Context, key string) error
...
 }
 
-// GetByKey gets the object identified by the given key from the store and returns the same as a runtime.Object.
+// GetByKey gets the object identified by the given key from the store and returns the same as
+// a runtime.Object. Three outcomes:
+//   - The key holds a live object -> returns the typed runtime.Object.
+//   - The key was previously deleted -> returns [minkapi.ErrObjectDeleted]
+//     so callers can distinguish "tombstoned" objects from non-existent ones.
+//   - The object for the key was never created -> returns apierrors.NotFound.
 func (s *InMemResourceStore) GetByKey(ctx context.Context, key string) (o runtime.Object, err error) {
```
elankath, line 196: Please add another test in `store_test.go` for tombstoned objects.

api/minkapi/types.go
```diff
@@ -67,20 +67,21 @@ type ResourceStore interface {
...
 	Update(ctx context.Context, mo metav1.Object, opts ObjectOptions) error
 	// DeleteByKey deletes an object from the store by its key.
 	DeleteByKey(ctx context.Context, key string) error
-	// Delete deletes an object from the store by its name.
-	Delete(ctx context.Context, objName cache.ObjectName) error
-	// DeleteObjects deletes objects matching the given criteria and returns the count of deleted objects.
-	DeleteObjects(ctx context.Context, c MatchCriteria) (delCount int, err error)
+	// Delete deletes an object from the store by its name using the given [ObjectOptions].
+	Delete(ctx context.Context, objName cache.ObjectName, opts ObjectOptions) error
+	// DeleteObjects deletes objects matching the given criteria, using given [ObjectOptions]
```
aaronfern: Just for clarity
```suggestion
	// DeleteObjects deletes a list of objects matching the given criteria, using given [ObjectOptions]
```

minkapi/view/store/store.go
```diff
@@ -65,15 +66,18 @@ func NewInMemResourceStore(args *minkapi.ResourceStoreArgs) *InMemResourceStore
...
-// Reset resets the backing cache for this story and re-initializes the watch broadcasters.
+// Reset resets the backing cache (clearing both live entries and tombstones) and the watch
+// broadcaster for this store.
 func (s *InMemResourceStore) Reset() error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	s.cache = cache.NewStore(cache.MetaNamespaceKeyFunc)
+	s.cache = cache.NewStore(cache.DeletionHandlingMetaNamespaceKeyFunc)
 	return nil
 }
```
aaronfern: The doc string says "reset cache and broadcaster", but only the cache is being reset. From context, I assume only the cache in intended to be reset, and you might have to change the docstring

minkapi/view/store/store.go
```diff
@@ -336,6 +363,69 @@ func (s *InMemResourceStore) DeleteObjects(ctx context.Context, c minkapi.MatchC
 	return
 }
 
+// markDeleted records a tombstone for the given key. If a live object existed at the key it
+// is removed from the cache, a Deleted watch event is broadcast and a [cache.DeletedFinalStateUnknown]
+// marker is parked at the same key in the underlying cache.Store.
+//
+// markDeleted returns apierrors.NotFound when the key is already tombstoned or never existed.
+func (s *InMemResourceStore) markDeleted(ctx context.Context, key string) error {
```
aaronfern, line 402: Can you please add some unit tests for this?

minkapi/view/sandbox.go
```diff
@@ -230,11 +242,19 @@ func (v *sandboxView) ListMetaObjects(ctx context.Context, gvk schema.GroupVersi
 	if err != nil {
 		return
 	}
-	if myMax >= delegateMax {
-		maxVersion = myMax
-	} else {
-		maxVersion = delegateMax
+	// Filter out delegate items whose name has been tombstoned in this sandbox's store.
+	s, _ := v.GetResourceStore(gvk)
```
aaronfern: Why ignore a possible error here?

minkapi/view/sandbox.go
```diff
@@ -230,11 +242,19 @@ func (v *sandboxView) ListMetaObjects(ctx context.Context, gvk schema.GroupVersi
...
+		filtered := delegateItems[:0]
+		for _, item := range delegateItems {
+			if _, gerr := s.GetByKey(ctx, objutil.CacheName(item).String()); errors.Is(gerr, minkapi.ErrObjectDeleted) {
+				continue
+			}
+			filtered = append(filtered, item)
+		}
+		delegateItems = filtered
 	}
+	maxVersion = max(myMax, delegateMax)
```
aaronfern: iiuc correctly, delegateMax is the highest object resource version present in the list. After filtering out items, do we not have to reverify this value?

minkapi/view/store/store.go
r4mek: Will fix it.

minkapi/view/store/store.go
r4mek: Added in the latest commit.

minkapi/view/sandbox.go
r4mek: Code has changed with the latest commit. Please have another look.

minkapi/view/sandbox.go
r4mek: Code has changed with the latest commit. Please have another look.

minkapi/view/base_test.go
```diff
@@ -248,7 +268,8 @@ func TestCombinePrimarySecondary(t *testing.T) {
 			},
 		},
 	}
-	combined := combinePrimarySecondary(primary, secondary)
+	tombtoneKeys := sets.Set[string]{}
```
takoverflow:
```suggestion
	tombstoneKeys := sets.Set[string]{}
```

minkapi/view/sandbox.go
```diff
@@ -226,16 +238,16 @@ func (v *sandboxView) ListMetaObjects(ctx context.Context, gvk schema.GroupVersi
...
-	delegateItems, delegateMax, err := v.delegateView.ListMetaObjects(ctx, gvk, criteria)
+	sandboxTombstonedKeys, err := listTombstonedKeys(ctx, v, gvk)
 	if err != nil {
 		return
 	}
-	if myMax >= delegateMax {
-		maxVersion = myMax
-	} else {
-		maxVersion = delegateMax
+	delegateItems, delegateMax, err := v.delegateView.ListMetaObjects(ctx, gvk, criteria)
```
takoverflow, line 232: After getting the `sandboxTombstonedKeys`, I would expect this to be passed as `ExcludeNames` to the `deletegateView` list call.

```suggestion
	criteria.ExcludeNames = sandboxTombstonedKeys
	delegateItems, delegateMax, err := v.delegateView.ListMetaObjects(ctx, gvk, criteria)
```

This allows for no changes to be done to `combinePrimarySecondary`, WDYT @r4mek

minkapi/view/store/store.go
```diff
@@ -430,7 +459,7 @@ func (s *InMemResourceStore) markDeleted(ctx context.Context, key string) error
 // Accepts both value and pointer forms for robustness.
 func isTombstone(obj any) bool {
 	switch obj.(type) {
-	case cache.DeletedFinalStateUnknown, *cache.DeletedFinalStateUnknown:
```
takoverflow: Is this removal of check for `*cache.DeletedFinalStateUnknown` intentional?

Because the docstring still refers to checking both value and pointer forms.

r4mek: >Is this removal of check for *cache.DeletedFinalStateUnknown intentional?

Yes. As we will never have a case where the object is `*cache.DeletedFinalStateUnknown`.

api/minkapi/types.go
```diff
@@ -67,24 +67,27 @@ type ResourceStore interface {
...
+	Delete(ctx context.Context, objName cache.ObjectName, opts ObjectOptions) error
+	// DeleteObjects deletes a list of objects matching the given criteria, using given [ObjectOptions]
+	// and returns the count of deleted objects.
+	DeleteObjects(ctx context.Context, c MatchCriteria, opts ObjectOptions) (delCount int, err error)
 	// List lists objects matching the given criteria.
 	List(ctx context.Context, c MatchCriteria) (listObj runtime.Object, err error)
 	// ListMetaObjects lists metadata objects matching the given criteria.
 	ListMetaObjects(ctx context.Context, c MatchCriteria) (metaObjs []metav1.Object, maxVersion int64, err error)
+	// ListTombstonedKeys lists the keys of tombstoned objects.
+	ListTombstonedKeys(ctx context.Context) (tombstoneKeys sets.Set[string], err error)
```
aaronfern:
```suggestion
	ListTombstonedKeys(ctx context.Context) (tombstonedKeys sets.Set[string], err error)
```

api/minkapi/types.go
```diff
@@ -243,9 +248,12 @@ type App struct {
 }
 
 // MatchCriteria defines criteria for matching Kubernetes objects.
+// ExcludeNames takes precedence over Names.
```
aaronfern: Instead of just specifying one values precedence over another, it might be more meaningful to explain the precedence chain in the docstring.
For eg: `The precedence of criteria for matching is as follows in the order of highest precedence to lowest: Namespace, ExcludeNames, Names, LabelSelector`

minkapi/view/sandbox.go
```diff
@@ -303,34 +322,72 @@ func (v *sandboxView) GetWatcher(ctx context.Context, gvk schema.GroupVersionKin
...
+		v.changeCount.Add(1)
+		return nil
+	} else if errors.Is(gerr, minkapi.ErrObjectDeleted) {
+		// Already tombstoned: NotFound, mirroring K8s API server behavior for repeated delete.
+		return apierrors.NewNotFound(schema.GroupResource{Group: gvk.Group, Resource: gvk.Kind}, objName.String())
+	} else if !apierrors.IsNotFound(gerr) {
+		return gerr
+	}
+	// Sandbox doesn't have it; consult delegate.
+	delegateObj, err := v.delegateView.GetObject(ctx, gvk, objName)
```
aaronfern, line 335: Do we have to check if delegateObj is a tombstone?

r4mek: Currently we don't have a case where we can have tombstones in `delegateView`. But later there might be a scenario where tombstones can be present in `delegateView`.

WDYT @elankath ?

r4mek: As per our discussions, since the error return wraps `apierrors.NotFound` as well as `minkapi.ErrObjectDeleted` in case of tombstone. So this should not cause any issue.

minkapi/view/sandbox.go
```diff
@@ -156,26 +156,25 @@ func (v *sandboxView) GetObject(ctx context.Context, gvk schema.GroupVersionKind
...
 	objName := objutil.CacheName(obj)
 	sandboxObj, err := v.getSandboxObject(ctx, gvk, objName)
-	if err != nil && errors.Is(err, minkapi.ErrObjectDeleted) {
-		// Tombstoned: caller is updating something that has been deleted from the sandbox.
-		return apierrors.NewNotFound(schema.GroupResource{Group: gvk.Group, Resource: gvk.Kind}, objName.String())
-	}
-	if err != nil && !apierrors.IsNotFound(err) {
-		return err
+	if err != nil {
+		if errors.Is(err, minkapi.ErrObjectDeleted) {
```
takoverflow: This change is not needed now since `GetByKey` already wraps it with `NotFound`

So just this should be fine
```go
if err != nil && !apierrors.IsNotFound(err) {
	return err
}
```

r4mek: I would still need to check for `minkapi.ErrObjectDeleted`, right?

```go
if err != nil && (errors.Is(err, minkapi.ErrObjectDeleted) || !apierrors.IsNotFound(err)) {
	return err
}
```

takoverflow: Yes, I misinterpreted the case when if its `ErrObjectDeleted` wrapped with `NotFound`, you'll not want to fetch the object from the delegate. Thanks for clarifying!

minkapi/view/store/store_test.go
```diff
@@ -581,6 +869,116 @@ func TestWatch(t *testing.T) {
 	}
 }
 
+func TestGetWatcher(t *testing.T) {
+	tests := map[string]struct {
```
elankath: We already have watcher table driven test. Please give better name for this test if it cannot be merged into the table driven test

r4mek: Will remove the test. Later we can add if needed.

